PR #34, opened 2014-12-30
Added "hasEncryptedAttributes"

relaxnow:
Added "hasEncryptedAttributes" to check whether an assertion contains encrypted attributes.

Refactor private 'encryptedAttribute' to plural and simplify isNameIdEncrypted check.

### DIFF
--- a/src/SAML2/Assertion.php
+++ b/src/SAML2/Assertion.php
@@ -49,11 +49,11 @@ class SAML2_Assertion implements SAML2_SignedElement
     /**
      * The encrypted Attributes.
      *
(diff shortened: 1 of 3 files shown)


relaxnow:
@jaimeperez could you review?

relaxnow:
@jaimeperez ?

thijskh:
Encrypted attributes are discouraged by saml2int. So I don't think SSP will give a high priority to do something with this.

Nonetheless I think there's not much harm to merge this as the library does support it and the method is simple, tested and might be useful to someone.

PR #36, opened 2015-01-30
Modified the xsDateTime regex

paukenba:
according to SAML v2.0 core specification's paragraph 1.3.3 and xmlschema-2#dateTime I think the Z can be omitted

### DIFF
--- a/src/SAML2/Utils.php
+++ b/src/SAML2/Utils.php
@@ -683,7 +683,7 @@ class SAML2_Utils
         $matches = array();
 
         // We use a very strict regex to parse the timestamp.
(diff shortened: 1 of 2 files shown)


relaxnow:
I remembered reading something about that stupid Z and here it is:
https://groups.google.com/forum/#!msg/simplesamlphp/7m07m9gLzzA/Alck5HnmVCcJ

Sorry, looks like this is a bug in the spec and we need to err on the side of requiring the Z to make sure it's not local time (which we can't convert).

relaxnow:
Thank you very much for your contribution by the way and cool to get to know another ssp/saml2 user :).

thijskh:
Indeed, there's also an "official" statement that omitting the `Z` is not something that should be supported: https://lists.oasis-open.org/archives/saml-dev/201310/msg00001.html

thijskh:
I did merge the test you added though.

paukenba:
thank you guys, and thanks for this great lib

PR #38, opened 2015-05-06
Fix parse time

newmen:
Not all service providers generates IssueInstant value in the Authn Request as SAML2 accepts.
For example, Odin [Service Automation](http://www.odin.com/products/automation/) generates request with timezone shift. Microsoft Azure accepts requests with shifted timezone but SAML2 not yet.


thijskh:
Wouldn't it be better to fix those SP's that produce the wrong format rather than to make SSP more forgiving? Afterall, being more forgiving means those wrong formats proliferate and incompatibility ensures..

Afterall, the standard says: "All SAML time values have the type xs:dateTime [...] and MUST be expressed in UTC form, with no timezone component." where "UTC form" means the form which specifies `Z`.

newmen:
Okay
No problem

PR #41, opened 2015-05-11
Add support for sender-vouches SubjectConfirmation Method.

brianV:
This patch adds a new constant which references the sender-vouches SubjectConfirmation Method.


brianV:
The constant added in this pull request is needed by https://github.com/simplesamlphp/simplesamlphp/pull/200. This was implemented in response to simplesamlphp#50 (https://github.com/simplesamlphp/simplesamlphp/issues/50)

PR #42, opened 2015-05-22
Use robrichards/xmlseclibs instead simplesamlphp/xmlseclibs

Maks3w:
Current version have the patches provided by simplesamlphp fork


Maks3w:
This change need to be released as stable tag before change the dependency on simplesamlphp/simplesamlphp

thijskh:
Thanks!

PR #44, opened 2015-06-25
Scoping: idplist attributes configurable

reinier-vegter:
At this moment we're not able to set custom attributes on samlp:IDPEntry, sometimes needed to implement IDP scoping.
This change is backwards compatible, and gives possibility to provide attributes per IDPEntity (see https://simplesamlphp.org/docs/stable/simplesamlphp-scoping).

Please accept the pull request.
Thx.

Reinier

### DIFF
--- a/src/SAML2/AuthnRequest.php
+++ b/src/SAML2/AuthnRequest.php
@@ -731,7 +731,14 @@ class SAML2_AuthnRequest extends SAML2_Request
                 $idplist = $this->document->createElementNS(SAML2_Const::NS_SAMLP, 'IDPList');
                 foreach ($this->IDPList as $provider) {
                     $idpEntry = $this->document->createElementNS(SAML2_Const::NS_SAMLP, 'IDPEntry');
-                    $idpEntry->setAttribute('ProviderID', $provider);
+                    if (is_string($provider)) {
+                        $idpEntry->setAttribute('ProviderID', $provider);
+                    }
+                    elseif (is_array($provider)) {
+                        foreach ($provider as $attribute => $value) {
+                            $idpEntry->setAttribute($attribute, $value);
+                        }
+                    }
                     $idplist->appendChild($idpEntry);
                 }
                 $scoping->appendChild($idplist);
--- a/src/SAML2/AuthnRequest.php
+++ b/src/SAML2/AuthnRequest.php
@@ -733,8 +733,7 @@ class SAML2_AuthnRequest extends SAML2_Request
                     $idpEntry = $this->document->createElementNS(SAML2_Const::NS_SAMLP, 'IDPEntry');
                     if (is_string($provider)) {
                         $idpEntry->setAttribute('ProviderID', $provider);
-                    }
-                    elseif (is_array($provider)) {
+                    } elseif (is_array($provider)) {
                         foreach ($provider as $attribute => $value) {
                             $idpEntry->setAttribute($attribute, $value);
                         }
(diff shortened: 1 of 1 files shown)


reinier-vegter:
See http://blogs.technet.com/b/ad/archive/2015/02/11/how-to-use-azure-ad-to-land-users-on-their-custom-login-page-from-within-your-app.aspx

Nice post about SAML + Azure / AD.
For example, for Azure you don't only need the ProviderID attribute, but also the Name attribute.

If accepted, maybe it's also possible to already merge this one back to the master branch, as this line of code is the same accross branches, but thats up to you guys.

BTW, thx for the good work on this awesome project!

reinier-vegter:
Config example:

    'IDPList' => array(
      array(
        'ProviderID' => 'https://my-provider.com',
        'Name' => 'my-provider.com',
      ),
    ),

relaxnow:
@reinier-vegter thank you for your contribution!
Would you mind fixing the PHPCS error and adding a unit test?

relaxnow:
Could you also fix the phpdoc: https://github.com/reinier-vegter/saml2/blob/master/src/SAML2/AuthnRequest.php#L392 ?

I looked at the contents and it looks BWC and get's my :+1: once these issues have been fixed.
@jaimeperez you agree?

reinier-vegter:
Hi @relaxnow , 

I extended phpdoc a bit, and fixed a tiny bug (scoping wasn't added at all in this branch :) ).
Phpcs errors/warnings were already fixed, but for unittest I don't have the time right now.

BTW, I didn't find any test where Scoping was covered at all.

Reinier

relaxnow:
Thanks!
I know we have a testing deficit right now, that's why it's important to not decrease coverage with new code.
But if @jaimeperez agrees this can be added I can add a unit test.

relaxnow:
@jaimeperez ?

jaimeperez:
Hi guys!

Thanks a lot for your contribution Reinier, and sorry so much for the time it took me to take a look at it. I agree that this is a valuable addition and I am definitely positive about merging your PR. However, apart from the missing tests, I see one potential issue.

The problem is that while the old code was assuming that the values in the array were always a `ProviderID`, and therefore the attribute was fixed to that one, the new code does not make any assumptions, allowing the user in practice to set **any** attribute to the `IDPEntry` element. That would generate SAML messages that are not compliant with the schema:

```xml
<element name="IDPEntry" type="samlp:IDPEntryType"/>
<complexType name="IDPEntryType">
   <attribute name="ProviderID" type="anyURI" use="required"/>
   <attribute name="Name" type="string" use="optional"/>
   <attribute name="Loc" type="anyURI" use="optional"/>
</complexType>
```

I think the simplest way to avoid this is to define an array with the possible attribute names defined by the schema for this element. Then, for every attribute parsed in the configuration, verify if the key in the configuration array is contained in the array with all the valid attribute names. If it is, include it in the `IDPEntry`. If not, just skip it. This way we avoid errors generated in the proxy about non-conforming SAML.

How does it sound?

reinier-vegter:
@jaimeperez, good point, thanks for pointing to the schema.
Fixed and tested.

thijskh:
Good; I really prefer to see a test though.

Because the issue has been lingering so long I've just gone ahead and added the test myself rather than waiting much longer.

Thanks for your contribution, @reinier-vegter.

jaimeperez:
Thanks for merging this and adding the tests @thijskh!

reinier-vegter:
No problem, and thanks for merging @thijskh !

relaxnow:
I just found a bug caused by this PR because now the scoping element is only added if there is an IDPList (not for instance when there is only a RequesterID).

I'm sorry I didn't catch this, I should've reviewed the code after the change and added a more comprehensive test for scoping (which I will do presently). :disappointed:

reinier-vegter:
Hi @relaxnow ,

I mistakenly introduced it with commit 0a936e9494969addbbe9317bfa29984d8390a3da (so called bugfix... :sob: ) .
Should be fixed by only reverting that commit.
Apologies

PR #45, opened 2015-07-13
[Vulnerability] Prevent XXE Processing

DRvanR:
Currently when using the SAML2 library without calling `libxml_disable_entity_loader(true);` prior to using the library you are vulnerable to [XXE Processing vulnerability][1]. This PR helps mitigating this vulnerability by adding defensive measures in the library as well as documenting external measures in he README.

[1]: https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Processing


DRvanR:
This was discovered by a security audit on a project using this library.

Please review and merge asap, as potentially anyone using this library is vulnerable.

/cc @jaimeperez, @relaxnow, @thijskh

thijskh:
Might be useful to note that simpleSAMLphp does call `libxml_disable_entity_loader()` (since January 2015) and OpenConext EB does so as well. Also, `libxml2` has disabled external entity loading by default since version 2.9.

jaimeperez:
Hi all!

Not very worried about this for the reasons Thijs was mentioning, but in any case, I think it's a good idea to be as careful as possible with the SAML2 library. The update looks good to me. @relaxnow, @thijskh, what do you think?

DRvanR:
@jaimeperez @thijskh I agree it is not an issue for simpleSAMLphp or OpenConext EB (we researched that before opening this PR) but for any other application using this library it can be a vulnerability, hence the request to merge asap.

relaxnow:
:+1: from me too, so merged. 
Thanks @DRvanR ! Excellent work as usual.

Still doesn't really help any existing application using this library with an old libxml (2.9.0, released 11 Sep 2012) and/or PHP (fixed in PHP 5.3.23, 5.4.13 released 14 Mar 2013).
Then again if you're on such an old unpatched PHP version you'll have bigger problems than a unlikely XXE deep in the SAML2 library.

PR #48, opened 2015-07-27
ECP support to saml2 library

sitya:
This patch adds ECP support to saml2 library. It is a reorganized version of (abandoned?) https://github.com/bkryza/saml2. I have fixed a bug in it, and merged to the latest source of simplesamlphp/saml2 repository.

I have tested and it works fine with a modified simplesamlphp code. If this PR can be merged I will make another PR to simplesamlphp/simplesamlphp repository with the working, ECP enabled code.

More about SAML2 ECP: https://wiki.oasis-open.org/security/SAML2EnhancedClientProfile


thijskh:
Hi sitya, thanks for your contribution!

As I'm really knowledgable about ECP myself I have to defer to someone else on the contents of the PR. However, I do notice that several blocks of code have been commented out. I suggest to either remove them or to add some explanation why it's useful to keep them.

Also, a unit test is definitely a must in my opinion.

sitya:
Thanks for the reply. I removed the comments remained in the code, and I will write a unit test as well.

sitya:
Ok, it is done, I hope, it could be merged.

jaimeperez:
Hi @sitya!

Finally I got some time again to work on this. Unfortunately, I'm not carrying good news. I've got the impression that we need a non-negligible amount of work yet to be done in order to merge this. I'll add some comments directly in the code here and in the `simplesamlphp/saml2` pull request, but in essence:

* When an authentication error happens, no SAML response is sent back to the ECP client, contrary to what the ECP specification says in the first paragraph of section 2.3.6:

> Regardless of the success or failure of authentication of the principal and of processing the `<samlp:AuthnRequest>` message, the identity provider **MUST** return a `<samlp:Response>` message or SOAP fault [...] using the SAML SOAP binding [SAML2Bind], with the `<samlp:Response>` message in the body.

* Section 3 of the ECP specification (*Conformance*), and in particular section 3.1.1, identifies the features and functionalities that *must* be supported in order to claim conformance with the ECP profile. The PR barely complies with any of them:

> An identity provider that conforms to this profile MUST adhere to the relevant normative text in Section 2.3, including the **verification of channel bindings** and the use of **"Holder of Key" subject confirmation**. The use of **X.509 certificates as a proof mechanism MUST be supported**. Other key forms are OPTIONAL.
> 
> It MUST **support the use of HTTP Basic Authentication**, **TLS Client Authentication**, and the **XML Signature mechanism described in section 2.3.4.1**.
> 
> It MUST also **support verification of channel bindings of type "tls-server-end-point"** [RFC5929] between itself and the client during authentication via signed message.

Fixing the issues we have already in the code will take time (unfortunately it won't be fast since I need time myself to write all comments down in the code as well as going through all of it, and my time is **very** limited right now). But we also need to add lots of other stuff, in particular the *Holder of Key* subject confirmation, *TLS Client Authentication* and the *XML Signature mechanism* mentioned in the spec. None of that is trivial, unfortunately.

I don't know how it is for you in terms of resources or whether you think all of this can be done fast, but I think it's not very realistic to think this will take less than a month, so I'm seriously considering to withdraw this PR from the 1.14.0 milestone, and move it to 1.15.0 (to be released sometime early next year, hopefully).

What do you think?

sitya:
Hi,

> On 2015. nov. 18., at 17:09, Jaime Pérez Crespo <notifications@github.com> wrote:
> 
> Finally I got some time again to work on this.

thanks.

> Unfortunately, I'm not carrying good news. I've got the impression that we need a non-negligible amount of work yet to be done in order to merge this. I'll add some comments directly in the code here and in the simplesamlphp/saml2 pull request, but in essence:
> 
> 	• When an authentication error happens, no SAML response is sent back to the ECP client, contrary to what the ECP specification says in the first paragraph of section 2.3.6:
> Regardless of the success or failure of authentication of the principal and of processing the <samlp:AuthnRequest> message, the identity provider MUST return a <samlp:Response> message or SOAP fault [...] using the SAML SOAP binding [SAML2Bind], with the <samlp:Response> message in the body.

Ok, it is trivial, I pushed the modified code. Actually I tought it sent SOAP error before, but it is clearer to send <samlp:Response>. So it works now.

> 	• Section 3 of the ECP specification (Conformance), and in particular section 3.1.1, identifies the features and functionalities that must be supported in order to claim conformance with the ECP profile. The PR barely complies with any of them:
> An identity provider that conforms to this profile MUST adhere to the relevant normative text in Section 2.3, including the verification of channel bindings and the use of "Holder of Key" subject confirmation. The use of X.509 certificates as a proof mechanism MUST be supported. Other key forms are OPTIONAL.
> 
> It MUST support the use of HTTP Basic Authentication, TLS Client Authentication, and the XML Signature mechanism described in section 2.3.4.1.
> 
> It MUST also support verification of channel bindings of type "tls-server-end-point" [RFC5929] between itself and the client during authentication via signed message.
> 
> Fixing the issues we have already in the code will take time (unfortunately it won't be fast since I need time myself to write all comments down in the code as well as going through all of it, and my time is very limited right now). But we also need to add lots of other stuff, in particular the Holder of Key subject confirmation, TLS Client Authentication and the XML Signature mechanism mentioned in the spec. None of that is trivial, unfortunately.
> 
> I don't know how it is for you in terms of resources or whether you think all of this can be done fast, but I think it's not very realistic to think this will take less than a month, so I'm seriously considering to withdraw this PR from the 1.14.0 milestone, and move it to 1.15.0 (to be released sometime early next year, hopefully).
> 
> What do you think?

I am not too familiar with HoK but I will check the SSP code as it has a solution for these requirements, and if it could be used by ecp module, I could extend the PR, otherwise yes, it would be more realistic to reassign this PR to 1.15. May I give a couple of days for checking this?

sitya:
>> I don't know how it is for you in terms of resources or whether you think all of this can be done fast, but I think it's not very realistic to think this will take less than a month, so I'm seriously considering to withdraw this PR from the 1.14.0 milestone, and move it to 1.15.0 (to be released sometime early next year, hopefully).

>> What do you think?

> I am not too familiar with HoK but I will check the SSP code as it has a solution for these requirements, and if it could be used by ecp module, I could extend the PR, otherwise yes, it would be more realistic to reassign this PR to 1.15. May I give a couple of days for checking this?

I have made some tests and HoK worked for me with ECP, but I am not sure if it covered every scenario. I used 
* a modified metadata
* a prepared `xml` with `SOAP envelop` and `AuthnRequest` with `HoK binding` inside and
* `curl` 

for the testing.

But `XML Signature mechanism` with verification of channel-binding is not supported now, and I am not sure, when I could be able to develop this feature.

All in all, as I see, ECP is pretty rarely used, but when it is used in most cases the capability of HTTP BasicAuth is enough. So I might suggest -- if there is no more bug -- to release a limited version which could satisfy most of the IdP/SP administrators who liked to support ECP with SSP.

src/SAML2/Binding.php
```diff
@@ -34,6 +34,8 @@ public static function getBinding($urn)
                 return new SAML2_HTTPRedirect();
             case SAML2_Const::BINDING_HTTP_ARTIFACT:
                 return new SAML2_HTTPArtifact();
+            case SAML2_Const::BINDING_SOAP_RESPONSE:
+                return new SAML2_SOAP();
```
jaimeperez: Why a new constant, specially if you are just creating a SAML2_SOAP object?

src/SAML2/Binding.php
```diff
@@ -75,7 +77,7 @@ public static function getCurrentBinding()
                     return new SAML2_HTTPPost();
                 } elseif (array_key_exists('SAMLart', $_POST)) {
                     return new SAML2_HTTPArtifact();
-                } elseif ($contentType === 'text/xml') {
+                } elseif (in_array($contentType, array('text/xml','application/soap+xml'))) {
```
jaimeperez: Wouldn't we need to add `application/vnd.paos+xml`, as per the specification?

src/SAML2/Const.php
```diff
@@ -33,11 +33,21 @@ class SAML2_Const
...
+
+    /**
      * The URN for the SOAP binding.
      */
     const BINDING_SOAP = 'urn:oasis:names:tc:SAML:2.0:bindings:SOAP';
 
     /**
+     * An internal binding that indicates that we should send a response by SOAP.
+     */
+    const BINDING_SOAP_RESPONSE = 'http://simplesamlphp.org/ns/bindings/SOAP-Response';
```
jaimeperez: This is the constant I was referring to before. We already have a `BINDING_SOAP` constant defining the proper namespace that identifies it. Bindings are defined for both requests and responses in the protocol (where applicable), so I don't see the need to define a specific constant carrying a custom namespace.

src/SAML2/SOAP.php
```diff
@@ -7,6 +7,9 @@
  */
 class SAML2_SOAP extends SAML2_Binding
 {
+
+    public $AssertionConsumerServiceURL;
```
jaimeperez: This is wrong, and derived from the confusion with the constant identifying the binding that I was commenting on. Every class that extends `SAML2_Binding` has a `location` property that stores the location where the message is being sent.

Adding an `$AssertionConsumerServiceURL` property here has two implications:

* We open up for bugs, like the one we have now that leads to the client to raise an exception, because the `AssertionConsumerServiceURL` is set to an empty string and that doesn't match the real ACS URL.
* We are assuming that this class is only suitable for SAML responses, since those are the ones that could carry an ACS URL, but the class is a generic `SAML2_SOAP` that doesn't match with either what it does or how it is used (i.e. the `BINDING_SOAP_RESPONSE` constant.

src/SAML2/SOAP.php
```diff
@@ -16,9 +19,15 @@ class SAML2_SOAP extends SAML2_Binding
...
     public function send(SAML2_Message $message)
     {
+        assert('!empty($this->AssertionConsumerServiceURL)');
         header('Content-Type: text/xml', TRUE);
         $outputFromIdp = '<?xml version="1.0" encoding="UTF-8"?>';
         $outputFromIdp .= '<SOAP-ENV:Envelope xmlns:SOAP-ENV="http://schemas.xmlsoap.org/soap/envelope/">';
+        $outputFromIdp .= '<SOAP-ENV:Header>';
+        $outputFromIdp .= '<ecp:Response xmlns:ecp="urn:oasis:names:tc:SAML:2.0:profiles:SSO:ecp" ';
+        $outputFromIdp .= '  SOAP-ENV:mustUnderstand="1" SOAP-ENV:actor="http://schemas.xmlsoap.org/soap/actor/next" ';
+        $outputFromIdp .= '  AssertionConsumerServiceURL="' . $this->AssertionConsumerServiceURL . '"/>';
```
jaimeperez: IMHO, this method should differentiate what kind of `SAML2_Message` is getting (*instanceof*), and act accordingly, instead of using an extra property that's not used and that leads to inconsistencies. Then, if this is a SAML response, the method should build this SOAP header with the `ecp:Response` inside, and add an `AssertionConsumerServiceURL` attribute, using the value located in `$this->destination`.

sitya:
Oh. I hoped that the source project I used as base for the PR is precise enough, but apparently it is not. Even it works. :( I will fix these bugs as well.

JohnMaguire:
Similar to the pull request simplesamlphp/simplesamlphp#233, I pushed a branch with namespaced references here: https://github.com/JohnMaguire/saml2/tree/sitya/master

It wasn't as necessary here because the autoloader had some magic built-in to help with classes prefixed `SAML2_` (`simplesamlphp` has the same thing for `SimpleSAML_`, but not for `SAML2_` which was the problem i was addressing), but may as well be updated.

JohnMaguire:
Actually, the reference to `SAML2_XML_ecp_Response` will fail without [my change](https://github.com/JohnMaguire/saml2/commit/f4815ba922259e491de121ea1a1a51728755d205#diff-8d8685ee843d1e073e4baf8542e0595fR32).

src/SAML2/SOAP.php
```diff
@@ -19,23 +17,24 @@ class SAML2_SOAP extends SAML2_Binding
...
+                    '<soap-env:Header/><soap-env:Body /></soap-env:Envelope>';
+
+        $doc = new DOMDocument();
+        $doc->loadXML($envelope);
+        $soapHeader = $doc->getElementsByTagNameNS('http://schemas.xmlsoap.org/soap/envelope/', 'Header');
+        $soapBody = $doc->getElementsByTagNameNS('http://schemas.xmlsoap.org/soap/envelope/', 'Body');
+
+        if ($message->toSignedXML()->localName === 'Response') {
+            $response = new SAML2_XML_ecp_Response();
+            $response->AssertionConsumerServiceURL = $this->getDestination();
```
JohnMaguire: `$this->destination` does not appear to be set. Not sure what the best way to set it off-hand is, but calling `$message->getDestination()` returns what we want.

JohnMaguire:
With all the aforementioned changes, plus removing the Signature element inside the Response element (but leaving the one in the Assertion) I was able to authenticate against Exchange Online.

I'm not sure if the signature is being generated incorrectly, or if Microsoft simply rejects it. I configured a Shibboleth 3 server and compared the ECP responses in order to make this determination.

JohnMaguire:
> Azure requires that encryption be turned off and that only assertions be signed.  It will also be necessary to set a name identifier precedence so that the ECP endpoint responds with a format of "urn:oasis:names:tc:SAML:2.0:nameid-format:persistent".

Found this note [here](https://wiki.shibboleth.net/confluence/display/IDP30/Office+365), so I'm pretty sure the auth failure with a signed response is expected. I'm following up with Microsoft to confirm, but that may take some time.

JohnMaguire:
@jaimeperez As @sitya stated, I'm not familiar with SPs that make use of Holder of Key confirmation, or TLS client authentication, or the XML signature mechanism of authentication mentioned. Like him, I'd really like to see this PR go through for the purposes of Exchange Online.

Shibboleth takes the approach of requiring the web server to handle this, and just takes the `REMOTE_USER` server variable at face value:

> With a typical IdP, authentication of clients is handled using a web form of some kind to collect a username and password. This approach is not appropriate for most ECP deployments and is not supported by the Shibboleth ECP extension.
> 
> Instead, some form of container- or web server-based authentication is required "in front" of the ECP endpoint, essentially the same way the "RemoteUser" login handler that ships with the IdP works. This could be basic authentication with LDAP, Kerberos, etc., or could even be client certificate authentication with TLS. Anything you can use to get REMOTE_USER set will work, but you will often need to set this up in parallel with your IdP's "standard" login process for use by browsers.
> 
> The actual exchange between the ECP client and the IdP is based on SOAP, and future ECP support in the IdP will provide for a wider range of authentication options using SOAP and HTTP, and will not require extra setup work; the same back-end verifiers will be used for both the ECP and Browser profiles.
> 
> Note that the above refers to the "basic" ECP use case; in the delegation case, the authentication to the IdP is done using a service's certificate (using TLS) and the user's SSO assertion is attached as a SOAP header.

Taken from https://wiki.shibboleth.net/confluence/display/CONCEPT/ECP#ECP-Authentication

Perhaps we could take this approach for non-"HTTP Basic" authentication requests, at least until someone has a need for implementing something more? Although, I'm not sure this would knock out this bit:

> It MUST also support verification of channel bindings of type "tls-server-end-point" [RFC5929] between itself and the client during authentication via signed message.

However, at the very beginning of the spec, I see:

> This specification updates the original profile by adding support for "Holder of Key" subject confirmation [SAML2HOK] and channel bindings [ChanBind]. These additions are optional from a deployment perspective, and are incorporated in a backward-compatible fashion for use with existing implementations when the new features are not used. Both features can be used independently or together, to strengthen the security of the profile.

Reading through section 2.3.2, we see that the server MAY respond to CB/HoK headers.

JohnMaguire:
I've begun some work on the [saml-ecp](https://github.com/JohnMaguire/saml2/tree/saml-ecp) branch of my fork. I'm working on adding tests for this code, as well as polishing up a couple little things I noticed.

I would really like to see this get merged, so any input that can be offered on what needs to be done would be awesome.

src/SAML2/SOAPClient.php
```diff
@@ -34,11 +64,11 @@ public function send(Message $msg, SimpleSAML_Configuration $srcMetadata, Simple
         $ctxOpts = array(
             'ssl' => array(
                 'capture_peer_cert' => true,
-                'allow_self_signed' => true
+                'allow_self_signed' => true,
```
JohnMaguire: @sitya Is there a reason this is set beyond testing/debugging purposes? It seems like this should be set to `false` at least by default.

src/SAML2/SOAPClient.php
```diff
@@ -113,19 +145,20 @@ public function send(Message $msg, SimpleSAML_Configuration $srcMetadata, Simple
...
-        // Add soap-envelopes
+        /* Add soap-envelopes */
         $request = $msg->toSignedXML();
-        $request = self::START_SOAP_ENVELOPE . $request->ownerDocument->saveXML($request) . self::END_SOAP_ENVELOPE;
+        $request = self::START_SOAP_ENVELOPE.$request->ownerDocument->saveXML($request).self::END_SOAP_ENVELOPE;
 
         Utils::getContainer()->debugMessage($request, 'out');
 
-        $action = 'http://www.oasis-open.org/committees/security';
-        $version = '1.1';
```
JohnMaguire: Any reason these are removed? They're still referenced on line 160, and it looks like they could replace the SOAPAction header added in the new code.

src/SAML2/SOAP.php
```diff
@@ -18,17 +19,24 @@ class SOAP extends Binding
...
+        
+        $envelope = '<soap-env:Envelope xmlns:soap-env="http://schemas.xmlsoap.org/soap/envelope/">'.
+                    '<soap-env:Header/><soap-env:Body /></soap-env:Envelope>';
+
+        $doc = new DOMDocument();
+        $doc->loadXML($envelope);
+        $soapHeader = $doc->getElementsByTagNameNS('http://schemas.xmlsoap.org/soap/envelope/', 'Header');
+        $soapBody = $doc->getElementsByTagNameNS('http://schemas.xmlsoap.org/soap/envelope/', 'Body');
+
+        if ($message->toSignedXML()->localName === 'Response') {
```
JohnMaguire: It looks like there's a new assumption here that if we are sending a `Response` down from this SOAP binding, then it must be an ECP response... is that accurate? I imagine this SOAP binding was used for some other sort of response previously...

JohnMaguire: A safer alternative may be to copy `SOAP.php` to `PAOS.php`, leaving the original alone, and removing [this hack in simplesamlphp](https://github.com/simplesamlphp/simplesamlphp/pull/233/files#diff-456d2e8c8b05c3e68f3ddf81766b1326R252).

JohnMaguire: Did a bit of digging, and it looks the SOAP binding today is used for HTTP Artifact binding. `ArtifactResolutionService.php` creates a new `SOAP` binding, receives the artifact request, and responds with the stored artifact.

I believe the stored artifact could be a "Response" as an `AuthnRequest` can come in with the HTTP Artifact binding... I think this could break HTTP Artifact binding.

JohnMaguire:
FWIW, I did confirm with Microsoft that signing the Response is expected to cause an authentication failure. When using ECP with Exchange Online, only the Assertion may (and should) be signed.

thijskh:
Replaced by #96.

PR #49, opened 2015-09-01
Function libxml_disable_entity_loader is used to prevent XXE attacks

sgomez:
This PR is related with PR #45 and with simplesamlphp/simplesamlphp#241 issue. Now ```libxml_disable_entity_loader``` is enabled before to read XML data instead of leaving this responsibility to the user.

This PR fix a issue with ```SAML2_DOMDocumentFactory::fromFile()``` method, because if the entity loader is disabled the method ```DOMDocument::load``` is disabled too, launching an exception.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,10 +61,6 @@ Example:
 Security
 --------
 
-* **Ensure that before calling any code from this library you have called** `libxml_disable_entity_loader(true);`
-  this is required to prevent the [XXE Processing Vulnerability](https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Processing),
-  see also [this websec.io page](http://websec.io/2012/08/27/Preventing-XEE-in-PHP.html)
-
 * Should you need to create a DOMDocument instance, use the `SAML2_DOMDocumentFactory` to create DOMDocuments from
   either a string (`SAML2_DOMDocumentFactory::fromString($theXmlAsString)`), a file (`SAML2_DOMDocumentFactory::fromFile($pathToTheFile)`)
   or just a new instance (`SAML2_DOMDocumentFactory::create()`)
--- a/src/SAML2/AuthnRequest.php
+++ b/src/SAML2/AuthnRequest.php
@@ -594,7 +594,7 @@ class SAML2_AuthnRequest extends SAML2_Request
     public function encryptNameId(XMLSecurityKey $key)
     {
         /* First create a XML representation of the NameID. */
-        $doc  = new DOMDocument();
+        $doc  = SAML2_DOMDocumentFactory::create();
         $root = $doc->createElement('root');
         $doc->appendChild($root);
         SAML2_Utils::addNameId($root, $this->nameId);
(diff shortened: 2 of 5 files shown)


DRvanR:
Hello, 

Disclaimer: I am not a maintainer of this project :smile: 

first of all thanks for the PR! Good catch on the DomDocument instantiation in the `AuthnRequest`! Reading the referenced issue it is clear that this should be done. However a lot has changed bar using `libxml_disable_entity_loader` inside the library. What is the rationale about all the other changes? Also, there are some backwards compatibility issues here. Don't know if that will affect other projects, but it doesn't seem necessary (the added `LIBXML` constants for instance).

Aside from the above, I think there is some room for improvement:
- Rather than duplicating a lot of logic in the `fromFile` method, that should just load the file (with proper error handling :wink:) and use `fromString` to create the DomDocument.
- I'd advise against putting too much checks and balances with regards to content in this class (e.g. [new lines 103-109][2]) as that was (is?) not the purpose of this class. It should only provide secure DomDocument creation, not be judgemental about the contents :wink:
- The libxml error handing deserves some love i.e. use `libxml_get_last_error()` to get the error, clear the buffer afterwards as well, there could be fatal errors as well, etc.
- the error-handling for DomDocument has been removed, this must be reinstated as it breaks BC (also removes the necessity of the separate `empty()` check on the string, which coincidentally misses a `trim` call :wink:).

Think the above would also help in getting the build to pass :)

Just my 0,02 :wink: 

[1]: https://github.com/simplesamlphp/saml2/blob/ec7e6457b4cf43ff97791a8963cf1a982cb13c12/src/SAML2/DOMDocumentFactory.php#L25
[2]: https://github.com/sgomez/saml2/blob/fix-prevent-xxe/src/SAML2/DOMDocumentFactory.php#L103-L109

sgomez:
Hi @DRvanR 

Thanks for the comments:

* Yes, I was thinking about it. But I preferred to respect the same structure. You are right and I'm going to change it.
* This code prevents the use of DOCTYPE to prevent the Quadratic Blowup Attack or Generic Entity Expansion. You have more info [here](http://phpsecurity.readthedocs.org/en/latest/Injection-Attacks.html#defenses-against-xml-entity-expansion).
* There are two ways to handing XML errors: with ```libxml_use_internal_errors()``` enabled or disabled. I'm enabling internal errors because is the way [to fetch error information on libxml](http://php.net/manual/en/function.libxml-use-internal-errors.php), so Error Control Operator "@" is not longer necessary (and should be avoided, it could hide another problem not related with libxml). Besides, the new code process all the error messages iterating ```libxml_get_errors()``` (not just the last).
* Error handling is not removed (as I explained). And ```trim``` will be added too. Thanks!

The build has failed because the NPath complexity is above the limit (there are to many ```if``` sentences). It is really necessary [to check if the argument is a object or not](https://github.com/simplesamlphp/saml2/blob/master/src/SAML2/DOMDocumentFactory.php#L18)?

thijskh:
Fixed through #52.

PR #51, opened 2015-09-28
ENHANCEMENT: Add ability to specify SoapClient proxy_host and proxy_port

madmatt:
When using SimpleSAMLphp, this allows you to specify `saml.SOAPClient.proxyhost` and `saml.SOAPClient.proxyport` in your `authsources.php` configuration, and these will be passed through to the underlying `SOAPClient` object that is created, allowing PHP to use a proxy for making SOAP calls.

This is essential in locked-down environments that need to make all external API calls via a proxy, rather than directly accessing the internet - the rest of the SAML flow works fine because we just redirect the user's browser, but this call will otherwise fail.

### DIFF
--- a/src/SAML2/SOAPClient.php
+++ b/src/SAML2/SOAPClient.php
@@ -91,6 +91,14 @@ class SAML2_SOAPClient
             'stream_context' => $context,
         );
 
+        if ($srcMetadata->hasValue('saml.SOAPClient.proxyhost')) {
+            $options['proxy_host'] = $srcMetadata->getValue('saml.SOAPClient.proxyhost');
+        }
+
+        if ($srcMetadata->hasValue('saml.SOAPClient.proxyport')) {
+            $options['proxy_port'] = $srcMetadata->getValue('saml.SOAPClient.proxyport');
+        }
+
         $x = new SoapClient(NULL, $options);
 
         // Add soap-envelopes


thijskh:
simpleSAMLphp already has a `proxy` configuration option: https://github.com/simplesamlphp/simplesamlphp/blob/master/config-templates/config.php#L808. Would it make sense to reuse that here instead of introducing new options?

madmatt:
I've just force-pushed up a fix for Travis/PHP linting.

Regarding the `proxy` configuration option, I didn't see that one before, it could be useful to use that; although that would mean SOAPClient would also now be dependent on SimpleSAMLphp's base configuration file, as well as the source/destination metadata which is already passed through.

tractorcow:
I agree with @madmatt 's observation on this; saml2 is a dependency of, but doesn't depend on simplesamlphp, so it probably shouldn't rely on that API.

Maybe this should be merged, and simplesamlphp should push the proxy based on those settings somehow? Perhaps there needs to be a mapping in that module (rather than this one) that maps the proxy config to saml.SOAPClient.proxyhost?

tractorcow:
It looks like I was wrong... the `SimpleSAML_Configuration` used here is from the simplesamlphp/simplesamlphp library, not simplesamlphp/saml2. :P Funny how it's not a composer dependency isn't it?

I need to better understand these libraries sorry. :)

madmatt:
Hi @thijskh,

Any thoughts on my earlier comment? Specifically, I'm hesitant to tie `SOAPClient` to the simpleSAMLphp configuration array given it is already being passed in the source and destination metadata - so your proposal would add a third connection point between the two modules.

In addition, your proposal would mean that you couldn't use a different proxy setup for two different authentication sources (e.g. if you are using simpleSAMLphp as an IdP bridge between say an Azure AD system and some internal SAML auth system, you may need to use a proxy only for the connection to Azure, not for any local connections).

The only failing check is the coverage check, but given this class currently doesn't have unit tests, and it would be a large refactoring to be able to add unit tests to it (because of how the `SOAPClient` object is initialised in the middle of other code, it's impossible to mock out), I don't think it's worth addressing that in this pull request, what do you think?

tractorcow:
:+1:

thijskh:
Alright, I'll merge it. Thanks for your contibution!

madmatt:
Thanks @thijskh :)

PR #52, opened 2015-10-02
Scope remote entity loading protection

DRvanR:
Alternative implementation based on https://github.com/simplesamlphp/saml2/pull/49


thijskh:
Seems like a sane approach :+1:

PR #55, opened 2015-12-11
Catch errors when inflating or base64 encoding a SAMLRequest

thijskh:
The file did not yet have any tests so I added some testcases including the inflation exception.

I could not find a good string that would make `base64_decode` actually fail (since it's invoked non-strict), so this exception is not covered by the testcase.

This should address OpenConext/OpenConext-engineblock#118

### DIFF
--- a/tests/SAML2/HTTPRedirectTest.php
+++ b/tests/SAML2/HTTPRedirectTest.php
@@ -38,6 +38,22 @@ class SAML2_HTTPRedirectTest extends PHPUnit_Framework_TestCase
     }
 
     /**
+     * Test parsing a signed authentication request.
+     * It does not actually verify the validity of the signature.
+     */
+    public function testSignedRequestParsing()
+    {
+        $request = 'SAMLRequest=nVLBauMwEP0Vo7sjW7FpKpJA2rBsoNuGOruHXhZFHm8EsuRqxtv27yvbWWgvYelFgjfvzbx5zBJVazu56enkHuG5B6TktbUO5VhYsT446RUalE61gJK0rDY%2F7qSYZbILnrz2ln2QXFYoRAhkvGPJbrtiv7VoygJEoTJ9LOusXDSFuJ4vdH6cxwoIEGUjsrqoFUt%2BQcCoXLHYKMoRe9g5JOUoQlleprlI8%2FyQz6W4ksXiiSXbuI1xikbViahDyfkRSM2wD40DmjnL0bSdhcE6Hx7BTd3xqnqoIPw1GmbdqWPJNx80jCGtGIUeWLL5t8mtd9i3EM78n493%2FzWr9XVvx%2B58mj39IlUaR%2FQmKOPq4Dtkyf4c9E1EjPtzOePjREL5%2FXDYp%2FuH6sDWy6G3HDML66%2B5ayO7VlHx2dySf2y9nM7pPprabffeGv02ZNcquux5QEydNiNVUlAODTiKMVvrX24DKIJz8nw9jfx8tOt3&RelayState=https%3A%2F%2Fbeta.surfnet.nl%2Fsimplesaml%2Fmodule.php%2Fcore%2Fauthenticate.php%3Fas%3DBraindrops&SigAlg=http%3A%2F%2Fwww.w3.org%2F2000%2F09%2Fxmldsig%23rsa-sha1&Signature=b%2Bqe%2FXGgICOrEL1v9dwuoy0RJtJ%2FGNAr7gJGYSJzLG0riPKwo7v5CH8GPC2P9IRikaeaNeQrnhBAaf8FCWrO0cLFw4qR6msK9bxRBGk%2BhIaTUYCh54ETrVCyGlmBneMgC5%2FiCRvtEW3ESPXCCqt8Ncu98yZmv9LIVyHSl67Se%2BfbB9sDw3%2FfzwYIHRMqK2aS8jnsnqlgnBGGOXqIqN3%2Bd%2F2dwtCfz14s%2F9odoYzSUv32qfNPiPez6PSNqwhwH7dWE3TlO%2FjZmz0DnOeQ2ft6qdZEi5ZN5KCV6VmNKpkrLMq6DDPnuwPm%2F8oCAoT88R2jG7uf9QZB%2BArWJKMEhDLsCA%3D%3D';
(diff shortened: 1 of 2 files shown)


tests/SAML2/HTTPRedirectTest.php
```diff
@@ -0,0 +1,81 @@
...
+        $this->setExpectedException('Exception', 'Error while inflating');
+        $hr = new SAML2_HTTPRedirect();
+        $request = $hr->receive();
+        PHPUnit_Framework_Error_Warning::$enabled = $oldwarning;
+    }
+
+    /**
+     * test handling of non-deflated data in samlrequest
+     */
+    public function testNoRequetOrResponse()
```
jaimeperez: Typo? :-)

thijskh: Typo fixed

jaimeperez:
Thanks for fixing the typo! I was going to merge it right away, but I see the build is failing... Do you have any idea on why?

thijskh:
`phpmd-ci: [exec] /home/travis/build/simplesamlphp/saml2/src/SAML2/HTTPRedirect.php:90	The method receive() has an NPath complexity of 216. The configured NPath complexity threshold is 200.`

It is explained tersely here: http://phpmd.org/rules/codesize.html The problem seems to be that adding two more if statement tripped the receive() method over a 'too complex' threshold.

We can solve this either by refactoring receive() to move functionalitiy into their own methods, if we agree that this is indeed too complex; or we up the value of the NPath complexity threshold in `tools/phpmd/ruleset.xml` if we believe the check to be too strict.

DRvanR:
the `receive` method seems to be more complex that necessary, the following is functionally the same but should have a lower NPath:

```php
$data = self::parseQuery();
if (array_key_exists('SAMLRequest', $data)) {
    $message = $data['SAMLRequest'];
} elseif (array_key_exists('SAMLResponse', $data)) {
    $message = $data['SAMLResponse'];
} else {
    throw new Exception('Missing SAMLRequest or SAMLResponse parameter.');
}

if (isset($data['SAMLEncoding']) && $data['SAMLEncoding'] !== self::DEFLATE) {
    throw new Exception('Unknown SAMLEncoding: ' . var_export($data['SAMLEncoding'], true));
}

$message = base64_decode($message);
if ($message === false) {
    throw new Exception('Error while base64 decoding SAML message.');
}

$message = gzinflate($message);
if ($message === false) {
    throw new Exception('Error while inflating SAML message.');
}

SAML2_Utils::getContainer()->debugMessage($message, 'in');
$document = SAML2_DOMDocumentFactory::fromString($message);
$xml      = $document->firstChild;
$message  = SAML2_Message::fromXML($xml);

if (array_key_exists('RelayState', $data)) {
    $message->setRelayState($data['RelayState']);
}

if (!array_key_exists('Signature', $data)) {
    return $message;
}

if (!array_key_exists('SigAlg', $data)) {
    throw new Exception('Missing signature algorithm.');
}

$signData = array(
    'Signature' => $data['Signature'],
    'SigAlg'    => $data['SigAlg'],
    'Query'     => $data['SignedQuery'],
);

$message->addValidator(array(get_class($this), 'validateSignature'), $signData);

return $message;
```

thijskh:
Yes, the method was written basically with theoretical support for multiple encodings, but the SAML standards specifies only one possibility there. So it indeed makes sense to me to reduce complexity that way. If ever SAML 2.1 is released that allows additional encodings, that can be implemented in a seperate method then.

thijskh:
Hmm. PHPMD now claims that the NPath complexity is even higher than before, at 288.

DRvanR:
grrr, why doesn't phpmd consider exits :(
It makes sense actually:

```php
$data = self::parseQuery();

// 3 paths possible...
if (array_key_exists('SAMLRequest', $data)) {
    $message = $data['SAMLRequest'];
} elseif (array_key_exists('SAMLResponse', $data)) {
    $message = $data['SAMLResponse'];
} else {
    throw new Exception('Missing SAMLRequest or SAMLResponse parameter.');
}

// 3 paths possible (9 in total)
if (isset($data['SAMLEncoding']) && $data['SAMLEncoding'] !== self::DEFLATE) {
    throw new Exception('Unknown SAMLEncoding: ' . var_export($data['SAMLEncoding'], true));
}

$message = base64_decode($message);
// 2 paths possible (18 in total)
if ($message === false) {
    throw new Exception('Error while base64 decoding SAML message.');
}

$message = gzinflate($message);
// 2 paths possible (36 in total)
if ($message === false) {
    throw new Exception('Error while inflating SAML message.');
}

SAML2_Utils::getContainer()->debugMessage($message, 'in');
$document = SAML2_DOMDocumentFactory::fromString($message);
$xml      = $document->firstChild;
$message  = SAML2_Message::fromXML($xml);

// 2 paths possible (72 in total)
if (array_key_exists('RelayState', $data)) {
    $message->setRelayState($data['RelayState']);
}

// 2 paths possible (144 in total)
if (!array_key_exists('Signature', $data)) {
    return $message;
}

// 2 paths possible (288 in total)
if (!array_key_exists('SigAlg', $data)) {
    throw new Exception('Missing signature algorithm.');
}

$signData = array(
    'Signature' => $data['Signature'],
    'SigAlg'    => $data['SigAlg'],
    'Query'     => $data['SignedQuery'],
);

$message->addValidator(array(get_class($this), 'validateSignature'), $signData);

return $message;
```

Since solving this would mean just relocating code that for now seems to be better to keep together, I'd advise to add a docblock to the function containing:
```
/**
 * @SuppressWarnings(PHPMD.NPathComplexity)
 */
```
That makes the checker not verify the NPath here. Could be solved at a later time if needed.

DRvanR:
@thijskh, alternatively, remove the check [here][2] as `base64_decode` can only return false when the [`strict` param is set to true][1], which by default it is not. This would drop the NPath to 144. Not sure if that is wanted though :wink: 

[1]: http://php.net/manual/en/function.base64-decode.php#refsect1-function.base64-decode-parameters
[2]: https://github.com/thijskh/saml2/blob/a59587e0a37278f88a3cd9f34f9927824bd56b1e/src/SAML2/HTTPRedirect.php#L105-L108

thijskh:
There's precedent in the library for adding the docbook tag so I've gone that way.
I think is this mergable now.

rjkip:
Looks good to me :+1:

PR #56, opened 2015-12-17
Always add the scoping element to the AuthnRequest.

relaxnow:
Adding just a RequesterID would cause the Scoping element to not get added.
Reverses commit 0a936e9494969addbbe9317bfa29984d8390a3da by @reinier-vegter from PR #44.


thijskh:
:+1:

PR #61, opened 2016-02-23
shibmd:scope regexp attribute should default to false and be always explicit

thijskh:
The regexp attribute of a Scope used to default to null when not set explicitly.

According to the spec, when omitted this should be interpreted as it being false. Therefore, I think the null default is not correct to have.

Also, we should always output an explicit regexp attribute, as recommended by the spec for document signing reasons: https://wiki.shibboleth.net/confluence/display/SC/ShibMetaExt+V1.0

### DIFF
--- a/src/SAML2/XML/shibmd/Scope.php
+++ b/src/SAML2/XML/shibmd/Scope.php
@@ -27,9 +27,9 @@ class Scope
     /**
      * Whether this is a regexp scope.
      *
-     * @var bool|null
+     * @var bool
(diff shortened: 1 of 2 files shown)


DRvanR:
Good change, might want to seperate `testMarshalling` into three different tests (as it now goes through three Arrange, Act, Assert cycles in one test), other than that :+1:

thijskh:
Thanks. I'll refactor the tests "later".

PR #63, opened 2016-03-15
Allow setting dynamic value for AllowCreate in NameIdPolicy 

AngelAlvarado:
This is a use case I faced while implementing SSO for the [SPNEGO](https://en.wikipedia.org/wiki/SPNEGO) Identity Provider. 

Due the SPNEGO's specification you cannot logout a SPNEGO user gracefully **and** when trying to login again(after the user logs out) the user will get an exception. Therefore, you need to send “allowcreate=false” in the SAML request token in order to let the SPNEGO IdP know that it is 'OK' to not refresh/try to authenticate the user in every SAML insertion.

I don't know any other IdP with these specifications but definitely it is a valid use case. 

This PR allows to set AllowCreate as a 'True' or 'False' in the toUnsignedXML method.

### DIFF
--- a/src/SAML2/AuthnRequest.php
+++ b/src/SAML2/AuthnRequest.php
@@ -714,8 +714,8 @@ class AuthnRequest extends Request
             if (array_key_exists('SPNameQualifier', $this->nameIdPolicy)) {
                 $nameIdPolicy->setAttribute('SPNameQualifier', $this->nameIdPolicy['SPNameQualifier']);
             }
-            if (array_key_exists('AllowCreate', $this->nameIdPolicy) && $this->nameIdPolicy['AllowCreate']) {
-                $nameIdPolicy->setAttribute('AllowCreate', 'true');
+            if (array_key_exists('AllowCreate', $this->nameIdPolicy) && $this->nameIdPolicy['AllowCreate'] === TRUE || $this->nameIdPolicy['AllowCreate'] === FALSE) {
+                $nameIdPolicy->setAttribute('AllowCreate', var_export($this->nameIdPolicy['AllowCreate'], TRUE) );
             }
             $root->appendChild($nameIdPolicy);
         }
--- a/src/SAML2/AuthnRequest.php
+++ b/src/SAML2/AuthnRequest.php
@@ -714,8 +714,9 @@ class AuthnRequest extends Request
             if (array_key_exists('SPNameQualifier', $this->nameIdPolicy)) {
                 $nameIdPolicy->setAttribute('SPNameQualifier', $this->nameIdPolicy['SPNameQualifier']);
             }
-            if (array_key_exists('AllowCreate', $this->nameIdPolicy) && $this->nameIdPolicy['AllowCreate'] === TRUE || $this->nameIdPolicy['AllowCreate'] === FALSE) {
-                $nameIdPolicy->setAttribute('AllowCreate', var_export($this->nameIdPolicy['AllowCreate'], TRUE) );
+            if (array_key_exists('AllowCreate', $this->nameIdPolicy) && $this->nameIdPolicy['AllowCreate'] === true
+                || $this->nameIdPolicy['AllowCreate'] === false) {
+                $nameIdPolicy->setAttribute('AllowCreate', var_export($this->nameIdPolicy['AllowCreate'], true));
             }
             $root->appendChild($nameIdPolicy);
         }


jaimeperez:
¡Hola Ángel!

As a rule of thumb, PRs need to pass the build in order for us to merge them. I see the main problem here is the coding standards. Could you please take a look at our [notes for developers](https://github.com/simplesamlphp/saml2/blob/master/README-DEV.md) and fix the build?

¡Gracias!

AngelAlvarado:
¡Hola Jaime!

My apologies for the missing CS. I took a look at the developer notes and fixed the code. 

Saludos.

jaimeperez:
¡Gracias!

Now that I'm looking a bit more into this, I don't think this is needed at all. The `AllowCreate` element defaults to `false` when not specified. This means unless `AllowCreate` is set in the `NameIDPolicy` of an `AuthnRequest` **and** is set to `true`, the `AllowCreate` element will be skipped in the resulting authentication request and therefore the default `false` value must be interpreted by the recipient of the request.

Therefore, your modification to SimpleSAMLphp should be enough for this to work. That doesn't mean we shouldn't allow specifying the `AllowCreate` element even when we want it to be `false` explicitly, but it's going to be easier for you if we don't need to update the SAML2 library.

Could you please verify if your modification to SimpleSAMLphp works **without** this PR?

AngelAlvarado:
Hi!

Ok, then there may be something wrong specifically with the simpleSAMLphp version 1.13.0 (or at least the SAML2 included in this [this build](https://simplesamlphp.org/res/downloads/simplesamlphp-1.13.0.tar.gz)). If you want to replicate this, I am using this [Vagrant box](https://github.com/AngelAlvarado/continuous-delivery-saml-sso/releases) which installs the version 1.13.0 and configures an IdP and an SP. It downloads the version 1.13.0 from the [Simplesamlphp.org's downloads](https://simplesamlphp.org/res/downloads/) **not** the repo. The box is only for **testing** purposes. 

Using a fresh installation of the 1.13.0 `AllowCreate` always defaults to **true**.  Using this patch [simplesamlphp/pull/347](https://github.com/simplesamlphp/simplesamlphp/pull/347/files) and configuring a SP like this:

>'openidp' => array( //we will named our SP 'openidp'
        'saml:SP', //we want to use SAML for this SP
        'entityID' => 'http://192.168.33.99/simplesaml/idp', //this is the ID for this SP, later on it will be used in the IdP
        'privatekey' => 'saml.pem', //stored in simplesaml/cert
        'certificate' => 'saml.crt', //stored in simplesaml/cert
        'AllowCreatePolicy' => FALSE,
        'NameIDPolicy' => 'urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress',
    ),

This is what the SAML token looks like, notice AllowCreate disappears:
>    < samlp:NameIDPolicy Format="urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress" / >

I'm not sure what's the SAML2 library's version included in [this build](https://simplesamlphp.org/res/downloads/simplesamlphp-1.13.0.tar.gz) - maybe v0.5.0?- but **1.x**, **2.x** and **master** have this chunk of code in src/SAML2/AuthnRequest.php:

`if (array_key_exists('AllowCreate', $this->nameIdPolicy) && $this->nameIdPolicy['AllowCreate']) {
$nameIdPolicy->setAttribute('AllowCreate', 'true');`

Which won't let set `AllowCreate` to false. I may be doing something wrong here, I am sure.

AngelAlvarado:
Using:

```
git clone -b simplesamlphp-1.13 https://github.com/simplesamlphp/simplesamlphp.git
cd simplesamlphp 
php composer.phar install
```
I have the same behavior described in the last comment. `AllowCreate` always is set to true and applying patch `Allowcreate` disappears.

jaimeperez:
¡Hola!

>Using a fresh installation of the 1.13.0 `AllowCreate` always defaults to **true**.  Using this patch [simplesamlphp/pull/347](https://github.com/simplesamlphp/simplesamlphp/pull/347/files) and configuring a SP like this:

>```php
'openidp' => array( //we will named our SP 'openidp'
        'saml:SP', //we want to use SAML for this SP
        'entityID' => 'http://192.168.33.99/simplesaml/idp', //this is the ID for this SP, later on it will be used in the IdP
        'privatekey' => 'saml.pem', //stored in simplesaml/cert
        'certificate' => 'saml.crt', //stored in simplesaml/cert
        'AllowCreatePolicy' => FALSE,
        'NameIDPolicy' => 'urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress',
    ),
```

>This is what the SAML token looks like, notice AllowCreate disappears:
>```xml
<samlp:NameIDPolicy Format="urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress" />
```

Yes, that's exactly what I was expecting.

> Which won't let set `AllowCreate` to false. I may be doing something wrong here, I am sure.

No, you are completely right. The code does not allow setting `AllowCreate` to false indeed, but the point is that **you don't need to**, because **the absence of `AllowCreate` is equivalent to `AllowCreate="false"`**. If the IdP is following the specification, it should just work like that.

AngelAlvarado:
Qué hay @jaimeperez 

Thanks for replying. Now I understand the absence of `AllowCreate`. I'm curious ,where is this specified? Is there any documentation regarding this? Or is this based on IdP case-basis?

jaimeperez:
¡Hola!

It's in the SAML standard, part of the [*core* document](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf). In section **3.4.1.1 Element \<NameIDPolicy\>**, you can read:

>2123   `AllowCreate` [Optional]
2124       A Boolean value used to indicate whether the identity provider is allowed, in the course of fulfilling the
2125       request, to create a new identifier to represent the principal. **Defaults to "false". When "false", the**
2126       **requester constrains the identity provider to only issue an assertion to it if an acceptable identifier for
2127       the principal has already been established**. Note that this does not prevent the identity provider from
2128       creating such identifiers outside the context of this specific request (for example, in advance for a
2129       large number of principals).

(bold is mine)

jaimeperez:
Resolved by recent commits. This is not strictly necessary for simplesamlphp/simplesamlphp#347 to work, but it makes sense anyway.

PR #64, opened 2016-03-17
Allow setting dynamic value for AllowCreate in NameIdPolicy in 1.x

AngelAlvarado:
@jaimeperez. Related to [pull#63](https://github.com/simplesamlphp/saml2/pull/63). This is for the 1.x branch. Let me know if it is ok to create two PRs for this.

### DIFF
--- a/src/SAML2/AuthnRequest.php
+++ b/src/SAML2/AuthnRequest.php
@@ -708,8 +708,9 @@ class SAML2_AuthnRequest extends SAML2_Request
             if (array_key_exists('SPNameQualifier', $this->nameIdPolicy)) {
                 $nameIdPolicy->setAttribute('SPNameQualifier', $this->nameIdPolicy['SPNameQualifier']);
             }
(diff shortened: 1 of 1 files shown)


thijskh:
This changes the behaviour for AllowCreate slightly.

saml2 does not specify what it expects as a value for AllowCreate, the code comment contains:
```
     * Set the NameIDPolicy.
     *
     * This function accepts an array with the following options:
     *  - 'Format'
     *  - 'SPNameQualifier'
     *  - 'AllowCreate'
     *
     * @param array $nameIdPolicy The NameIDPolicy.
```

In OpenConext-engineblock, we set AllowCreate to the string `"true"`. This is probably based on the implicit assumption that AllowCreate needs the value that will be put into the attribute. The addition of an `is_bool` check in this PR, makes that it is now always required to be an actual boolean.

We can change this and the interface is probably more logical this way. I think we should need to update the code comment block to make the interface more explicit to users.
@jaimeperez

PR #65, opened 2016-03-27
Fix KeyLoader::loadKeys

cb8:
This patch fixes a typo in `SAML2\Certificate\KeyLoader`.


thijskh:
Good catch, and thanks for including a test!

PR #76, opened 2016-09-07
Handle EPTI attribute parsing correctly

DRvanR:
As an eduPersonTargetedID is [specified as always containing a NameID][1], we can detect and handle that case by treating the attribute values as NameID. This also resolves the issue of an EPTI attribute not being serializeable, as reported [here][2] and discussed in #60.

Do note that this is a BC break and therefor should result in the release of a new major, albeit it arguably can be treated as a bugfix. It the latter case it could be released in the 2.x range.

[1]: http://software.internet2.edu/eduperson/internet2-mace-dir-eduperson-201310.html#eduPersonTargetedID
[2]: https://groups.google.com/forum/#!msg/simplesamlphp/Zn9KFqflUjU/6yhjPPOeBQAJ

### DIFF
--- a/src/SAML2/Assertion.php
+++ b/src/SAML2/Assertion.php
@@ -153,10 +153,21 @@ class Assertion implements SignedElement
     private $AuthenticatingAuthority;
 
     /**
-     * The attributes, as an associative array.
+     * The attributes, as an associative array, indexed by attribute name
      *
-     * @var array multi-dimensional array, indexed by attribute name with each value representing the attribute value
-     *            of that attribute. This value is an array of \DOMNodeList|string|int
+     * To ease handling, all attribute values are represented as an array of values, also for values with a multiplicity
+     * of single. There are 4 possible variants of datatypes for the values: a string, an integer, an array or
+     * a DOMNodeList
+     *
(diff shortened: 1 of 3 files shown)


DRvanR:
@jaimeperez @thijskh thoughts?

thijskh:
Yes, I believe this is what we discussed. We concluded that since two of the saml2 using projects already ran into this, it makes sense to do something about it in the library. And because we are certain that an EPTI is a nameID, it's an acceptable tradeoff to special case this attribute.

jaimeperez:
Fully agreed, big +1 here.

With this, I'll save myself the work to write a filter in SimpleSAMLphp to do this conversion automatically.

DRvanR:
Alright, shall we consider this to be a bugfix or a BC-break? I'd prefer it to be a bugfix, and release it in the 2.x range. This will prevent incompatibilities in consuming projects due to it being released as a new major.

jaimeperez:
That's a bit more complicated to answer. Breaking backwards compatibility in a bugfix is really bad, but having a new major seems a serious overkill...

I'd probably go for a new bugfix release in the 2.x branch, but those are just my 2 cents.

thijskh:
Yes, 2.x branch is fine.

jaimeperez:
Thanks **a lot** for this @DRvanR!

PR #77, opened 2016-09-13
Merge for release 2.3

DRvanR:


DRvanR:
@thijskh @jaimeperez seeing you two did the work here :wink: are there any commits that should not be merged for release 2.3?

thijskh:
I'm wondering if we should merge #69 also.

jaimeperez:
The only commit of those I perpetrated that could be a bit more problematic is [this one](https://github.com/simplesamlphp/saml2/pull/77/commits/00e38f85b417be1e10a2d738dd2f5ea82edb472c), since it bumps the minimum version of SimpleSAMLphp required by the compat layer.

thijskh:
Otherwise I'm :+1:

DRvanR:
#69 can and should be merged i.m.o. - a deprecation notice is still fully bc and just informative to users about what functionality will be removed at some point. 

The SSP logger is indeed technically a BC break, but only if you use the provided container with SSP which in and of itself is [not even a requirement][1] - seems that can and should be fixed with documentation and perhaps an additional interface (for dependecy inversion). Not sure whether or not breaking something that doesn't work out of the box and won't work without knowledge about SSP itself is considered a BC break...

I'm actually in favor of keeping both in as neither should be an issue. Thoughts?

[1]: https://github.com/simplesamlphp/saml2/blob/1cea30c465afbdf55551f05709e1a7cf4d83aaab/composer.json#L11-L18

thijskh:
Yes, agreed to add/keep both in.

jaimeperez:
I'm assuming both you and us are the main users of this library. So if you don't see a problem with keeping both, and I don't see it either... go ahead! 😉

DRvanR:
Awesome, thanks!

PR #78, opened 2016-09-14
Correct EPTI attribute value parsing

DRvanR:
backport of #76 to the release-1.x branch


DRvanR:
Also had to take some additional tests with it, but that's only for the better.

Thoughts @thijskh @jaimeperez ?

src/SAML2/Assertion.php
```diff
@@ -1462,6 +1493,16 @@ private function addAttributeStatement(DOMElement $root)
                 $attribute->setAttribute('NameFormat', $this->nameFormat);
             }
 
+            if ($name === SAML2_Const::EPTI_URN_MACE || $name === SAML2_Const::EPTI_URN_OID) {
+                foreach ($values as $eptiValue) {
+                    $attributeValue = $document->createElementNS(SAML2_Const::NS_SAML, 'saml:AttributeValue');
+                    $attribute->appendChild($attributeValue);
+                    SAML2_Utils::addNameId($attributeValue, $eptiValue);
```
jaimeperez: This merge had, unfortunately, two side effects:

* The first one is that this line is expecting you to encode the contents of an `eduPersonTargetedID` **always** as an array with some specific indexes. This breaks backwards-compatibility, because if you are creating an assertion manually (instead of parsing it from the original XML) and call `toXML()`, it will subsequently call `addAttributeStatement()`, which will then expect you to encode the attribute's value in a very specific way. Even worse, if the attribute's value is not an array, it will break completely with a white screen of death, since the second parameter of `SAML2_Utils::addNameId()` is declared as an array.

* On the other hand, this was tagged as `v1.10`, meaning everyone updating their installation via composer with a version requirement of `~1.0` or `^1.0` will install this new version, which could break their deployment. SimpleSAMLphp is an example of this, of course, and I've just received reports that people installing SSP 1.14.8 as an IdP and updating with composer are having trouble when trying to send an `eduPersonTargetedID` attribute.

We need to address this, in this branch and in 2.0. Master has it solved currently [thanks to the last commits](https://github.com/simplesamlphp/saml2/blob/master/src/SAML2/Assertion.php#L1513), since now an object is used and therefore `addNameId()` and `parseNameId()` are no longer necessary, but the code needs to avoid making any assumptions, since the attribute value can be provided by external parties.

Once this is fixed, we need a new release (v1.10.1? v1.11?) so that people can update and they can get their code working again (specially critical for SimpleSAMLphp acting as an IdP).

jaimeperez: It should now be resolved in all branches (master, 1.x & 2.x). Could you please verify that the fix doesn't break anything else?

The fix in 1.x and 2.x is a bit different than in master. Basically, because in master it is a much more complete fix, using the `SAML2\XML\saml\NameID` class instead of an array. However, this breaks backwards-compatibility too, so we need to evaluate it carefully.

DRvanR: @jaimeperez due to a broken hand I was not really capable of typing, hence a delayed response (hope to start work again tomorrow).

The proposed/implemented fixes are actually introducing potential bugs. As referenced in #76 , the value of an EPTI must be a NameID, and from #76 onward therefor expects the value to be a NameID compatible value. More specifically, that means at the very least an array with at least the key `Value` and as value of that key the actual value of the NameID. Therefor, allowing any arbitrary value as is done [here][1] is actually not conform the specification. Now, I'm not necessarily against the current implementation, but I do want to highlight that this allows for incorrect assertions to be created. 

[1]: https://github.com/simplesamlphp/saml2/blob/master/src/SAML2/Assertion.php#L1519-L1524

PR #79, opened 2016-10-12
Additional attributes on ContactPerson Element

tdiscuit:
In order for federation participants to claim compliance for [Sirtfi](https://wiki.refeds.org/display/SIRTFI/Guide+for+Federation+Participants), they need to define the security contact for their organization. The contactType of "other" is used, but Sirtfi requires additional attributes, `xmlns:remd` and `remd:contactType`, on the ContactPerson element ([shown here](https://wiki.refeds.org/display/STAN/Security+Contact+Metadata+Extension+Schema)).

Currently, this isn't possible, but this PR adds an associative array, $ContactPersonAttributes, which will add attributes to the ContactPerson element.

### DIFF
--- a/src/SAML2/XML/md/ContactPerson.php
+++ b/src/SAML2/XML/md/ContactPerson.php
@@ -95,7 +95,7 @@ class ContactPerson
         $this->EmailAddress = self::getStringElements($xml, 'EmailAddress');
         $this->TelephoneNumber = self::getStringElements($xml, 'TelephoneNumber');
 
-        foreach($xml->attributes as $attr) {
+        foreach ($xml->attributes as $attr) {
             if ($attr->nodeName == "contactType")
                 continue;
 
--- a/src/SAML2/XML/md/ContactPerson.php
+++ b/src/SAML2/XML/md/ContactPerson.php
@@ -64,6 +64,13 @@ class ContactPerson
     public $TelephoneNumber = array();
 
     /**
+     * Extra attributes on the contact element.
+     *
+     * @var array
+     */
+    public $ContactPersonAttributes = array();
+
+    /**
      * Initialize a ContactPerson element.
      *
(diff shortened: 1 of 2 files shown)


relaxnow:
@tdiscuit could you please add a unit test?

tdiscuit:
@relaxnow I've added unit tests and `SAML2\XML\md\ContactPerson` is 100% covered.

jaimeperez:
Thanks so much @tdiscuit!

I was wondering... would it make sense to have this as a new class, extending `ContactPerson`? Basically, a class specific to this use case, setting whatever needs to be set to comply with the SIRTFI contact, so that you just need to mark it like that in some way and then use the new class instead of the generic one. I think it could be worth evaluating the idea since `ContactPerson` is quite generic and the SIRTFI stuff is very specific on the contrary.

tdiscuit:
Anytime @jaimeperez! But I still have more work to do since I'll be submitting a PR for SimpleSAMLphp as well ;) ... once I hash that bit out.

I was thinking the same thing, but I went with this route since I am doing this for Sirtfi now, but there may be other trust frameworks that I don't know about that require additional attributes on other contact types. So by keeping it generic, it becomes very easy to implement. Of course I am open to discussion around that.

jaimeperez:
That's also true, although I don't know if there are other frameworks doing the same already. Maybe @joostd or @pmeulen know better?

In any case, I think this is also a sensible approach since the implementation is then extremely simple.

thijskh:
I'm not aware of any other trust frameworks that use extensions to the `ContactPerson` element. Likely the current approach is indeed just fine and suffices for now. To keep it 'agile' we can refactor this later should other use cases pop up. All in all I propose to merge this PR.

thijskh:
Merged. thanks!

jaimeperez:
Thank you guys!

PR #80, opened 2016-11-09
Fix 'Undefined namespace prefix' error

jeremyfrench:
I was trying to parse a SAML packet that had `saml2` as the SAML namespace prefix rather than `saml`. This caused a 'Undefined namespace prefix' error. 

It looks like the SAML namespace is declared by ssp as as 'saml_assertion' but for some reason it was not used with this Xpath. Changing it removes the error.


jaimeperez:
Thanks a lot @jeremyfrench! Good that you spotted this!

PR #81, opened 2016-11-29
Security update for signature validation

artem-diverta:
The underlying function "openssl_verify" returns -1 on error, which may happen (as happened for me while doing security checks) when the identity provider certificate is tampered with (for example when I substituted it with a DSA key)
In that case, -1 being true, well...

### DIFF
--- a/src/SAML2/Utils.php
+++ b/src/SAML2/Utils.php
@@ -165,7 +165,7 @@ class Utils
         }
 
         /* Check the signature. */
-        if (! $objXMLSecDSig->verify($key)) {
+        if ($objXMLSecDSig->verify($key) !== 1) {
             throw new \Exception("Unable to validate Signature");
         }
     }


pmeulen:
Hi artem-diverta,

Thank you for your PR, we are working on it.

pmeulen:
I can reproduce the issue. This PR fixes it. 
For the SAML usecase the certificate used for signature verification should not be under control of an attacker, which limits exploitability of this issue. There may be other scenarios where openssl_verify() returns an int < 0 in case of failure. I could not find such a scenario that would be easily exploitable by an attacker.

As this is a security related fix. Dependent libraries are encouraged to update.

@artem-diverta Thanks again for bringing this to our attention.

@thijskh OK to merge?

jaimeperez:
Hi everybody!

Thanks a lot @artem-diverta for notifying us about this issue and providing a fix for it. We have assessed the vulnerability and considered it to be moderate, due to the difficulty of exploitation.

New versions of the SAML2 library have been released, as well as version [1.14.10 of SimpleSAMLphp](https://github.com/simplesamlphp/simplesamlphp/releases/tag/v1.14.10). A [security advisory (SSPSA 201612-01)](https://simplesamlphp.org/security/201612-01) summarizing our findings has also been published.

Once again, thanks a lot for your help.

thijskh:
We could consider to check for 0 or -1 separately so we can give a different error message for each case.

Issue #82, opened 2016-12-07
Signature required even for HTTP-Redirect Response


artem-tim:
Hello,
I stumbled upon what appears to be a bug to me, please correct me if I'm wrong.
According to the specs, the AuthnResponse is not required to carry the signature inside the SAML message itself in the case of HTTP-Redirect (where the signature can be done via the query parameters). Actually, an IDP implemented with this library does precisely that : when generating an AuthnResponse using Redirect binding, the SAMLResponse body is left unsigned while the signature correctly appears among query parameters (comparing HTTPRedirect.php and HTTPPost.php).
The problem is, when such Response is treated by the lib, the Response/Processor isn't happy if there is no signature either at root or in assertions, whereas this response does comply with the specification and carries the signature in the query parameters.
I didn't start a discussion over a pull request because I'm not sure which is the best solution, but the ones I am thinking about are : 

1. Simply remove the whole block `if (!$response->isMessageConstructedWithSignature())` in Response/Processor::verifySignature. The HTTPRedirect adds the correct validator from query params to the Message so if it's invalid there will be an Exception anyway (I checked)
2. In a similar manner, remove the check in Response/Processor::verifySignature
3. If there is a real need to keep the verifications above, add a method (Message.php) to check if there is a validator based on function name ("validateSignature"), and add it as requirement to be unhappy along one of the verifications above.

Anyway, even though HTTP-Post is more common because of the length that can attain the encoded SAML message, in my humble opinion HTTP-Redirect deserves some love too - but maybe I'm missing something here.

Best regards,
Artem

pmeulen:
Hi Artem,

Sending SAML a Response via HTTP-Redirect would be unusual, it is not defined in the Web-SSO profile which states "Either the HTTP POST, or HTTP Artifact binding can be used to transfer
the message to the service provider through the user agent" (see: https://docs.oasis-open.org/security/saml/v2.0/saml-profiles-2.0-os.pdf). So I'm not surprised that the code currently assumes that a response is signed using XML-DSIG, as this is what these other bindings use.

The HTTP-Redirect binding is indeed defined for a SAMLResponse in https://docs.oasis-open.org/security/saml/v2.0/saml-bindings-2.0-os.pdf so it can be implemented based on that definition.

I'm curious as to why you want to do this? Do you have an IdP that sends responses using a HTTP-Redirect binding?

artem-tim:
Hi Pieter,

[...]

PR #83, opened 2016-12-08
Bug : Allowing a HTTPRedirect response to be processed

artem-tim:
A solution to the issue explained in #82


thijskh:
Closing as per discussion in #82

PR #85, opened 2016-12-19
Make explicit that AllowCreate must be bool

thijskh:
This is a followup to #63 and #64.

In those PR's, a new test was added on whether AllowCreate is a boolean, if not, it would be silently ignored. Making the typing more strict is a good thing, but in my opinion silently ignoring it if it has a different type is not right, and also not that it happens when building the request, not directly when using the setter. Also, it was never documented explicitly what the lib expected to get.

There's precedent for using `assert()` to enforce incoming variable types, but I believe throwing an exception is better. YMMV.

@jaimeperez @pmeulen @arothuis

### DIFF
--- a/tests/SAML2/AuthnRequestTest.php
+++ b/tests/SAML2/AuthnRequestTest.php
@@ -345,4 +345,168 @@ AUTHNREQUEST;
 
         $this->assertEquals($requesterId, $authnRequest->getRequesterID());
     }
+
+    /**
+     * Test getting NameIDPolicy
+     */
+    public function testGettingNameIDPolicy()
+    {
+        $xml = <<<AUTHNREQUEST
+<samlp:AuthnRequest
+  xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol"
+  xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion"
+  ID="_306f8ec5b618f361c70b6ffb1480eade"
+  Version="2.0"
+  IssueInstant="2004-12-05T09:21:59Z"
+  Destination="https://idp.example.org/SAML2/SSO/Artifact"
+  ProtocolBinding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Artifact"
+  AssertionConsumerServiceURL="https://sp.example.com/SAML2/SSO/Artifact">
+  <saml:Issuer>https://sp.example.com/SAML2</saml:Issuer>
+  <samlp:NameIDPolicy
+    AllowCreate="true"
+    SPNameQualifier="https://sp.example.com/SAML2"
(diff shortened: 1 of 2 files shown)


tests/SAML2/AuthnRequestTest.php
```diff
@@ -345,4 +345,167 @@ public function testRequesterIdIsReadCorrectly()
...
+     */
+    public function testSettingNameIDPolicy()
+    {
+        // basic AuthnRequest
+        $request = new AuthnRequest();
+        $request->setIssuer('https://gateway.example.org/saml20/sp/metadata');
+        $request->setDestination('https://tiqr.example.org/idp/profile/saml2/Redirect/SSO');
+        $request->setIssueInstant( Utils::xsDateTimeToTimestamp('2004-12-05T09:21:59Z'));
+
+        $nip = array("Format" => "urn:oasis:names:tc:SAML:2.0:nameid-format:transient",
```
arothuis: I'd suggest writing this out to reduce cognitive load: `$nameIdPolicy`

tests/SAML2/AuthnRequestTest.php
```diff
@@ -345,4 +345,167 @@ public function testRequesterIdIsReadCorrectly()
...
+
+        $this->assertEqualXMLStructure($expectedStructure, $requestStructure);
+
+        $this->assertXmlStringEqualsXmlString($expectedStructure->ownerDocument->saveXML(), $requestStructure->ownerDocument->saveXML());
+    }
+
+    /**
+     * Test setting NameIDPolicy with invalid type for AllowCreate.
+     */
+    public function testSettingNameIDPolicyToIncorrectTypeAllowCreate()
```
arothuis: You are only checking for one illegal type. What happens if I throw in an integer, object, resource, array?

We often use a [dataprovider method](https://github.com/SURFnet/Stepup-Middleware/blob/fe9716c6409a7dddaee1753160744da611e6705e/src/Surfnet/Stepup/Tests/Identity/Value/TimeFrameTest.php#L55-L69) (or [class](https://github.com/OpenConext/OpenConext-profile/blob/develop/src/OpenConext/Profile/Tests/DataProvider.php)) for this.

Not saying that it should be changed, but it would be nice to tackle possible cases of (implicit) type conversion this way.

tests/SAML2/AuthnRequestTest.php
thijskh: Fixed

tests/SAML2/AuthnRequestTest.php
thijskh: Yes, would be nice. Is quite a change, made a note in the comment and might implement this later (flw)

PR #88, opened 2017-01-18
Fix SOAPClient peer name verification when using a proxy

mtheunissen82:
Since PHP 5.6.x some openssl changes were introduced
which amongst others enable peer verification by default for all
encrypted client streams.
See: http://php.net/manual/en/migration56.openssl.php

There is a bug in Soapclient when used with a proxy it will verify
the wrong peer (the proxy) instead of the real destination.
See: https://bugs.php.net/bug.php?id=69137

This fix will allow a user to specify the correct peer_name of
the SoapClients stream_context in authsources.php
so the peer verification will be correct.

### DIFF
--- a/src/SAML2/SOAPClient.php
+++ b/src/SAML2/SOAPClient.php
@@ -88,6 +88,10 @@ class SOAPClient
             $ctxOpts['ssl']['cafile'] = $peerCertFile;
         }
 
+        if ($srcMetadata->hasValue('saml.SOAPClient.stream_context.ssl.peer_name')) {
+            $ctxOpts['ssl']['peer_name'] = $srcMetadata->getString('saml.SOAPClient.stream_context.ssl.peer_name');
+        }
+
         $context = stream_context_create($ctxOpts);
         if ($context === null) {
             throw new \Exception('Unable to create SSL stream context');


src/SAML2/SOAPClient.php
```diff
@@ -88,6 +88,10 @@ public function send(Message $msg, SimpleSAML_Configuration $srcMetadata, Simple
             $ctxOpts['ssl']['cafile'] = $peerCertFile;
         }
 
+        if ($srcMetadata->hasValue('saml.SOAPClient.stream_context.ssl.peer_name')) {
```
mtheunissen82: Add the ability to change the SSL peer_name stream_context. This is a workaround for [a bug](http://php.net/manual/en/migration56.openssl.php) where peer verification fails when using a proxy with SoapClient (ext/soap). The bug is present since [PHP 5.6.x where they updated the underlying openssl library](http://php.net/manual/en/migration56.openssl.php).

The fix has a low impact to the SAML2 library and as configuring a stream_context is a perfectly normal feature for any stream application it does no harm to the scope of the SAML2 library.

mtheunissen82:
Oh haha. Sorry, I am a github noob :). Thanks for the merge! In what release will it be expected to ship?

thijskh:
There's no concrete date for the next release of the saml2 library, sorry.

PR #89, opened 2017-01-27
* Add possibility to set `verify_peer` and `verify_peer_name` for the…

aWuttig:
… SOAPClient via authsource config (useful for development)

We need this really for development context when working with untrusted https certs.

### DIFF
--- a/src/SAML2/SOAPClient.php
+++ b/src/SAML2/SOAPClient.php
@@ -38,6 +38,14 @@ class SOAPClient
             ),
         );
 
+        if ($srcMetadata->hasValue('saml.SOAPClient.verify_peer')) {
+            $ctxOpts['ssl']['verify_peer'] = $srcMetadata->getBoolean('saml.SOAPClient.verify_peer');
+        }
+
+        if ($srcMetadata->hasValue('saml.SOAPClient.verify_peer_name')) {
+            $ctxOpts['ssl']['verify_peer_name'] = $srcMetadata->getBoolean('saml.SOAPClient.verify_peer_name');
+        }
+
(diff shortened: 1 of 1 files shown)


jaimeperez:
Hi @aWuttig!

I don't think this is a good idea. Doing so opens up for people using this in production with the security implications that would have.

If you need this for development, you could just modify the library to stop verifying certificates, and get rid of that modified library when deploying to production.

aWuttig:
But by default it is NOT set and the developer has to configure it with wisely knowing about the security issues!

jaimeperez:
I might be missing something (this is not my code and I may have used it a couple of times), but it is set:

https://github.com/simplesamlphp/saml2/blob/master/src/SAML2/SOAPClient.php#L86

Apart from that, the default for _verify_peer_ and _verify_peer_name_ is to do verify the other peer, at least from PHP 5.6. So in any case, the proper fix to do would be to unconditionally set both to `true`.

aWuttig:
howsoever... is there a way to override your classes from outside (maybe only in application context "development"). Otherwise i override your code on my local machine and this leads to modfied code unfortunatly :-1:

jaimeperez:
That was my first point. If this is done for development, what's the issue with modifying the code in your local machine? That's not what's going to end in production, right? So modifying the code in this library locally to fit your development environment shouldn't be a problem...

If you need to modify the code in production, that's a whole different story. But in that case, you should definitely not turn peer verification off, so unless you are running PHP < 5.6 and want to *enable* peer verification without specifying the certificate of the peer, I don't see the need for this PR. And even in that case, as I was saying, I'd rather not make it configurable, but if anything, enforce it so that we make sure peers are always verified regardless of what PHP version you are running.

PR #94, opened 2017-04-04
Issuer NameQualifier and explicit Format

falco76:
hi all,
this patch is useful to send Authn request compliant to SPID requirements (SAML2 italian governative AAI) and to use ssp as SPID SP
(SPID IDPs Require explicit Issuer format, NameQualifier and value in authn request)

<saml:Issuer NameQualifier="https://foo....."
                 Format="urn:oasis:names:tc:SAML:2.0:nameid-format:entity">https://foo...</saml:Issuer>

default behaviour is compliant to saml-core-2.0-os 8.3.6
i'm a newbie ... sorry


thijskh:
Hi Falco, Thanks for the PR. But I'm a bit confused by it.

I think `($this->Format === Constants::NAMEID_ENTITY)) || ($this->Format !== Constants::NAMEID_ENTITY)` will always evaluate to true. So what's the intention there?

Can you maybe for my understanding give two concrete examples of the XML 'before' and 'after' merging this commit.

As for the procedure, such changes should come with a unit test (for both scenarios of the setting). And please note that the automated build has failed due to some code style issues.

falco76:
hi thijskh,

the conditional code:
`  
if (
(($this->SAML2IssuerShowAll) && ($this->Format === Constants::NAMEID_ENTITY)) || 
($this->Format !== Constants::NAMEID_ENTITY)
)
`
Condition1=(($this->SAML2IssuerShowAll) && ($this->Format === Constants::NAMEID_ENTITY))
Condition2=($this->Format !== Constants::NAMEID_ENTITY) (original condition)
all conditions force to print extended attributes (namequalifier, format etc)


#### Before ####
a1. NO default format -> print extended attributes
`
<saml:Issuer NameQualifier="https://foo....."
Format="NO_DEFAULT_FORMAT">https://foo...</saml:Issuer>
`

a2. default format -> print basic  
`
<saml:Issuer>https://foo...</saml:Issuer>
`
i can't send namequalifier and default format


#### After ####
b1.SAML2IssuerShowAll=true and Format= default -> print extended attributes
`
<saml:Issuer NameQualifier="https://foo....." 
Format="urn:oasis:names:tc:SAML:2.0:nameid-format:entity">https://foo...</saml:Issuer>
`

b2.SAML2IssuerShowAll=false and Format= NO default -> print extended attributes
`
<saml:Issuer NameQualifier="https://foo....."
Format="NO_DEFAULT_FORMAT">  https://foo...</saml:Issuer>
`

b3.SAML2IssuerShowAll=true and Format= NO default -> print extended attributes
`
<saml:Issuer NameQualifier="https://foo....."
Format="NO_DEFAULT_FORMAT">https://foo...</saml:Issuer>
`

b4.SAML2IssuerShowAll=false and Format= default -> print basic
`
<saml:Issuer>https://foo...</saml:Issuer>
`

falco76:
@thijskh 
thanks for tips, now i have to learn phpunit ...

jaimeperez:
Hi!

Thanks for the PR @falco76!

To be honest, I'm not entirely sure I like the idea of breaking the standard. Is there a way we could fix this in some other way? @thijskh, what do you think?

falco76:
hi @jaimeperez,
this PR don't brake standard "by default" but only if you need ;) and if you want 

it (only) increase flexibility and permit use of saml2 library (and SSP) in SPID  and other (Eidas ?) SP scenarios,
see https://joinup.ec.europa.eu/sites/default/files/eidas_message_format_v1.0.pdf, 6.3 chapter

thank you a lot for these great projects

jaimeperez:
I concur. If we are merging this, tests are a must, especially to guarantee that the default behaviour is never broken.

thijskh:
@falco76 Any chance you can provide a test?

thijskh:
Thanks! I merged it now, with some small changes:
* Squashed all the fixup-commits into the first commit. Saves some history confusion.
* Renamed the testcase file to end in Test so phpunit will actually run it when running the whole suite.

PR #96, opened 2017-05-31
Add IdP support for ECP profile

JohnMaguire:
- Add constants for ECP
- Add support for ECP Response in SOAP binding
- Add support for SOAP in Binding::getBinding()

This is an updated version of #48 to fix some issues, and add unit tests. A PR to `simplesamlphp` is coming soon to complete support in the application.

### DIFF
--- a/src/SAML2/XML/ecp/Response.php
+++ b/src/SAML2/XML/ecp/Response.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace SAML2\XML\ecp;
+
+use DOMElement;
+use InvalidArgumentException;
+
+use SAML2\Constants;
+
+/**
+ * Class representing the ECP Response element.
+ */
+class Response
+{
+    /**
+     * The AssertionConsumerServiceURL.
+     *
+     * @var string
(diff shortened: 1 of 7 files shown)


JohnMaguire:
Added back a debug statement I accidentally removed.

src/SAML2/XML/ecp/Response.php
```diff
@@ -0,0 +1,77 @@
...
+
+use DOMElement;
+use InvalidArgumentException;
+
+use SAML2\Constants;
+
+/**
+ * Class representing the ECP Response element.
+ *
+ * @version $Id$
```
thijskh: There's a stray SVN Id tag here

thijskh:
Thanks a lot for this work. I think you mean this is a replacement for #48.

The build fails, maybe you want to add a comment
`@SuppressWarnings(PHPMD.ExitExpression)` to the `send()` method.

JohnMaguire:
Indeed, I meant #48. I have updated the commit with your suggestions, thanks!

Issue #97, opened 2017-06-05
PHP 7.1 support and xmlseclibs


chris1312:
`simplesamlphp/saml2` depends on `robrichards/xmlseclibs` and it is using mcrypt_* functions. Therefore, using this library it results into deprecated errors.

Why using Mcrypt is wrong? Read here https://paragonie.com/blog/2015/05/if-you-re-typing-word-mcrypt-into-your-code-you-re-doing-it-wrong

Has anybody tested SAML2 with `robrichards/xmlseclibs` 3.0.0+?

tvdijen:
It's only a couple of days old.....  Have you tested it for yourself?

thijskh:
Yes, I've done some testing with simpleSAMLphp and OpenConext-engineblock and their core functionality remains working, also their testsuites pass with this upgrade. We need to keep it at 2.0 or 3.0 as saml2 currently still supports PHP < 5.6.

chris1312:
That's nice to hear. Thanks for the update!

PR #98, opened 2017-06-05
Bump xmlseclibs version

chris1312:
#97

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "ext-dom": "*",
         "ext-zlib": "*",
 
-        "robrichards/xmlseclibs": "^2.0",
+        "robrichards/xmlseclibs": "^2.0|^3.0",
         "psr/log": "~1.0"
     },
     "require-dev": {


thijskh:
If you also bump the `composer.lock` then Travis can show us whether the testsuite still passes (spoiler: it does).

chris1312:
Sorry, I did not catch.. 

This project does not contain `composer.lock`. Do you say that you want me to commit `composer.lock` into the project? I see in Travis build (e.g. for PHP7) that the 3.0.0 version of `xmlseclibs` was installed.

thijskh:
Indeed, my mistake. The tests pass, that's a good start!

chris1312:
Is there anything more you expect from this pull request?

thijskh:
Not per se, just need to find some time to do some further testing, then merge.

jaimeperez:
Hi @chris1312!

I would also like to see how the discussion in robrichards/xmlseclibs#123 ends up before we trust the 3.0.0 to be stable and final. I'm afraid of bumping the version here to see days afterwards that it was removed to merge that PR instead.

That said, I'm really looking forward to this myself too...

chris1312:
OK, sure!

thijskh:
@jaimeperez Those two pr's are functionally the same, right? So would it matter much to change later?

jaimeperez:
Nah, I don't think there'll be changes to the code (not significantly anyway). I'm just afraid of some kind of mess if Rob tries to fix it so that proper credit is given to the code by merging the existing PR, but he doesn't seem to be responsive at all, and if something breaks, we can always fix it.

So given that we are in master here, 👍 from me too.

thijskh:
Thanks for your contribution!

Issue #90, opened 2017-01-31
Migrate assertions to something more modern


DRvanR:
The SAML2 library liberally uses the [`assert`][1] function as an assertion (sort of guard clause). Starting PHP7, this function has been changed to have a possible different implementation, that of [defining an expectation][2], with the ability to preserve the behavior of earlier versions.

Starting in PHP 7.2 the old behavior [will be deprecated][3]. Since PHP 5.6 is now [security support only][4] and PHP7.1 is considered the current stable version, SAML2 should start working on PHP7 compatibility.

My suggestion would be to remove all assert statements and replace them with as much typehints as possible and use [beberlei/assert][5] for any other assert statements.

[1]: https://github.com/simplesamlphp/saml2/search?utf8=%E2%9C%93&q=assert+path%3Asrc&type=Code
[2]: http://php.net/manual/en/function.assert.php#function.assert.expectations
[3]: https://wiki.php.net/rfc/deprecations_php_7_2#assert_with_string_argument
[4]: http://php.net/supported-versions.php
[5]: https://github.com/beberlei/assert

thijskh:
:+1:

jaimeperez:
Agreed.

I'm afraid we'll still be stuck for a while supporting older versions of PHP, but we also need to be fully compatible with newest versions anyway. I've always found asserts dangerous since they'll work only depending on your configuration, so I fully agree with using type hinting when possible and delegate further checks to an external dependency.

I took a look at [ihor/nspl](https://github.com/ihor/Nspl) some time ago indeed, but unfortunately, I found it just didn't work for me. Now I was trying it again and there seems to be some mess with the namespaces, and constants or function names are not found if you don't explicitly use the namespaces.

I've also taken a brief look at [beberley/assert](https://github.com/beberlei/assert) now, and I miss the possibility to have parameters that can be of several types (different than _null or..._). I'm not sure how much of those we have in the SAML2 library, but we have a lot of them in SimpleSAMLphp and I would definitely like to reuse the same dependency. The same problem affects [webmozart/assert](https://github.com/webmozart/assert), which is a fork of the previous and seems much more popular.

The only "reasonable" alternative I see is [wikimedia/assert](https://github.com/wikimedia/Assert), though the fact that they use the github repo as a mirror, that they have only released "betas", and that they are not actively developing it, makes me a bit afraid of using it, even though it looks much more flexible than the others...

DRvanR:
[...]

PR #110, opened 2017-12-01
Fix deprecated assert(...) usage

Trejjam:
Hi,

I replaced calling assert with string parameter, related to #90.

Of course it will be nice use strict types in method signature, but it is long term work.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,5 +36,10 @@
         "psr-0": {
             "SAML2\\": "tests/"
         }
+    },
+    "extra": {
+        "branch-alias": {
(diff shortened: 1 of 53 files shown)


thijskh:
Thanks!

thijskh:
Sorry, a bit too soon. The tests fail but this was swept under the carpet because they were broken anyway. But now that they are fixed, it shows that this PR has several test failures, so I reverted it. Can you investigate?

Trejjam:
Yes, I'll look at it

PR #111, opened 2017-12-09
Fix deprecated assert(...) usage

Trejjam:
Hi, I cannot reopen pull request https://github.com/simplesamlphp/saml2/pull/110, here is new version of it.


Trejjam:
Hi, I was trying make tests work on all supported versions of PHP. But support all of them looks like impossible task. On PHP 7.2 are broken dependencies (mostly bugs in depends libraries), and on the other hand nowadays version of dependent libraries are incompatible with old PHP 5.4 and 5.5. I was trying hack it but I was not completely successful.

Is is possible tag actual version with support of PHP 5.4 and move master on php>=5.6?

thijskh:
That is not possible at the moment as long as one of the major users of this library, simplesamlphp/simplesamlphp, has 5.4 as a minimum.

Trejjam:
It is actually possible. If you tag repository, then this release is used for old runtime, and master and its future release for other.
It is true, that this old runtimes are stack in past, but they already [are](http://php.net/supported-versions.php) ...

thijskh:
I understand that that is possible, but more branches to maintain in parallel means more work. This is not our top priority to spend our limited resources on to do that immediately (but of course we’ll upgrade at some point).

The PHP.net supported versions argument is not so relevant I think, 5.4 will be supported for years e.g. In RHEL.

thijskh:
My first objective is to reduce our existing use of legacy branches. When that's done for me is a good moment to start new branches. Of course others may have different priorities.

Trejjam:
I'll try make it work, but probably in 1. 2018

Trejjam:
I just over thinked it previously. Please take a look

I solved tests problem by updating phpunit, but it is also unsuported version. It works, but...

thijskh:
Thanks! Why is the change to NameIDTest necessary?

Trejjam:
This is part of a report from `ant phpunit-ci`:

```
     [exec] There was 1 error:
     [exec] 
     [exec] 1) SAML2\XML\md\NameIDTest::testToString
     [exec] PHPUnit_Framework_Exception: Could not load XML from object
     [exec] 
     [exec] /saml2/tests/SAML2/XML/saml/NameIDTest.php:65
     [exec] 
     [exec] FAILURES!
     [exec] Tests: 386, Assertions: 1282, Errors: 1.
```

I didn't search for real changes in the phpunit, but I think they stopped auto-convert anything into a string.

PR #114, opened 2018-02-23
Fix using old deleted classname

Ben-Ho:


tvdijen:
This one is a no-brainer, so I will merge it for now..
However, be aware that the 2.x branch is not supported any more and will not receive any (security-)updates. Please migrate to 3.x

Ben-Ho:
@tvdijen ok, thanks. I'll consider updating to 3.x.

Do you plan on creating a 2.3.6 tag as 2.3.5 introduced this problem for me and currently don't know if I can upgrade my dependent library to 3.x without forwarding incompatible changes. (for now i changed dependency in my library to exactly 2.3.4 instead of 2.3.*)

tvdijen:
Since we've introduced this issue only so recently, I guess a new tag is the right thing to do...

PR #120, opened 2018-05-01
Fix broken handling of unsigned response with encrypted assertions

jorissteyn:
The response processor currently assumes all assertions are decrypted,
and causes a fatal error when the response is unsigned:

    Call to undefined method SAML2\EncryptedAssertion::getWasSignedAtConstruction()

Whether the assertion is signed can only be determined after
decryption. This fix decrypts the assertions before checking if
the response or assertion is signed. In order to do this I had
to separate the verification from decryption in the assertion
processor.

### DIFF
--- a/src/SAML2/Assertion/Processor.php
+++ b/src/SAML2/Assertion/Processor.php
@@ -101,7 +101,7 @@ class Processor
     }
 
     /**
-     * @param \SAML2\Utilities\ArrayCollection|array $assertions Collection of decrypted assertions
+     * @param \SAML2\Utilities\ArrayCollection $assertions Collection of decrypted assertions
      *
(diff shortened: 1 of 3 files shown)


jorissteyn:
@pmeulen As discussed, I've pushed another commit that separates assertion validation from decryption, so we can first decrypt the assertion and throw the 'unsigned response' exception without first having to validate the assertion. This means the changes proposed in this PR do not change the behaviour of 'Response\Processor::process', the only behavioural change is to not get a fatal error when the response is unsigned and contains encrypted assertions.

If people use the assertion processor as a stand-alone component, they will have to update their code to reflect this change - that can't be done backwards compatibly - see e6c01fa. The build fails on coveralls, not fixing that now because we're not sure this should be merged.

PR #121, opened 2018-05-04
Fix method "hasEncryptedAttributes" of class "Assertion".

jbaron-gingco:


jbaron-gingco:
I don't think using `empty()` would be a good idea - it just hides too many things (i.e. `empty(0) === true`, `empty(0.0) === true`, `empty(0/0) === false`).

I agree with using this check at other places as well and will update the PR shortly.

jaimeperez:
As a general rule, yes. However, in this case, the values this variable can take are limited to `null` (the default) or an array, which might be empty, making it completely ok to use `empty()`. Bear in mind that the `encryptedAttributes` property is private, and it has no setter, so it's only our own code assigning values to it.

There's also an alternative: stop defaulting to `null`, and use `[]` as the default value instead. Then we can just test `$this->encryptedAttributes !== []`.

jaimeperez:
Ah, nevermind, I just saw you added a commit assuming an empty array as the default value.

Thanks again!

PR #106, opened 2017-11-08
support storing attributes as \SAML2\XML\saml\Attribute objects

dub357:
This is a fairly big change, but should be backwards compatible.  In conjunction with a change to simplesamlphp (for which another pull request will be generated), it will allow for finer control of attributes (specifically NameFormats and FriendlyNames) since simplesamlphp doesn't currently support attribute friendly names or different name formats per attribute

PR #122, opened 2018-05-11
support storing attributes as \SAML2\XML\saml\Attribute objects

tvdijen:
updated \SAML2\Assertion to use \SAML2\XML\saml\Attribute objects instead of an array of string|int|DOMElement.

This is a fairly major change, but the getAttributes and setAttributes methods have default params to handle backwards compatibility.
This is a "fix" for issue https://github.com/simplesamlphp/simplesamlphp/issues/690

This PR is a continuation of https://github.com/simplesamlphp/saml2/pull/106 and the credits for it should go to @dub357


thijskh:
I have a hard time to follow what this tries to accomplish and how it relates to HoK and SOAP.

tvdijen:
@dub357 Are you reading this?
I hope I haven't messed up reconstructing your original PR, but since the original commits are gone on your fork, there's no way for me to be sure..

dub357:
So the stuff related to bindings was actually not supposed to be part of the original PR.  My intent was to only have saml2:master pull in my changes to SAML2/Assertion.php and SAML2/XML/saml/AttributeValue.php.  The binding stuff was committed after my original PR was created and since I'm new to GitHub and didn't really understand how the pull request process works, I didn't realize any additional commits on the same branch were going to be added to the open PR.  

The binding changes were my attempt to make all bindings return their URN and should have probably not been committed/pushed to my default branch.  During my testing with other SPs, I've run into the situation where they expect a response to come back using the same saml2 web browser profile binding (request vs post) that the original authn request was sent with (which is really a bug in their code).  Because of this I was changing simplesamlphp to see if that could be 'fixed'.  It did solve the problem, but I didn't mean for that stuff to end up in the PR.  I attribute this to being a GitHub newbie LOL.  

I also see that during the copy/merging/squashing/whitespace removal nightmare tvdijen and I were stuck in, some of the changes didn't make it over because the binding stuff looks incomplete as there was also a change to SAML2\Binding.php (adding "abstract public function getURN();") that I don't see in the commit.  There may be other things that were missed, but like tvdijen stated, I can't tell now for some reason.  I have absolutely no idea how all these changes have been removed in my fork.  Its almost like my fork is even with simplesamlphp/saml2:master now as all the commits I pushed are gone now.  wtf.  I have no idea how that has happened.  Somehow it happened after I had to fix all the whitespace issues directly in the fork because I couldn't figure out how to pull changes from master back into my fork.  At least I still have the changes in my local repo...

Anyway - the intent for my original PR was only to have the Assertion object store an array of Attribute objects instead of scalar values and introduce a way to define name formats and friendly names per attribute.  Should have really only been the 2 files I mentioned above.

tvdijen:
@dub357 I've removed the unnecessary stuff now.
Could you please check this against your local repo and make sure nothing is missing?

dub357:
Looks like one other change is needed...not sure how I didn't catch this, maybe there isn't a test for it, but in SAML2/XML/saml/AttributeValue.php, the "\XML\saml\NameID" references are not correct.  They should probably just be "NameID" since that class is in the same namespace as AttributeValue.  Other than that, it looks OK to me.

tvdijen:
Okay, then it's up for review again!

dub357:
I see there has been no movement on this PR...its been like 6 months and nothing...no feedback, no approval, nothing.  What can we do to get this back on track?  

One of the biggest drawbacks with simplesamlphp is its limited ability to deal with attributes of assertions.  I want to add more capability to simplesamlphp in this area, but the biggest hurdle is this library and its lack of treating attributes as objects.   Folks have already pushed more and more code into the Assertion class in this library to 'control' attributes (overridding xsi types) when in all reality, none of that should really be in there at all.  Its API should give you the ability to get/set the attributes as objects so that the developer has more control to do what they need with them....whether that is getting/setting their friendly names, nameformats or xsi types.  
The code I wrote in this PR this is backwards compatible...any idea what the hold up is?

thijskh:
Just a lack of time with the regular contributors I'm afraid. We'll get to it.

tvdijen:
There's still a few conflicts left in Assertion.php that I don't know how to resolve

src/SAML2/Assertion.php
```diff
@@ -257,6 +263,18 @@ public function __construct(DOMElement $xml = null)
...
+        $this->attributes = array();
+        $this->attributesValueTypes = array();
+        $this->attributeNameFormats = array();
+        $this->attributeFriendlyNames = array();
+        $this->nameFormat = Constants::NAMEFORMAT_UNSPECIFIED;
+        $this->certificates = array();
+        $this->AuthenticatingAuthority = array();
+        $this->SubjectConfirmation = array();
+        $this->requiredEncAttributes = false;
+>>>>>>> updated \SAML2\Assertion to use \SAML2\XML\saml\Attribute objects instead of an array of string|int|DOMElement.
```
thijskh: PR contains several of these merge leftovers in the commits.

src/SAML2/XML/saml/AttributeValue.php
```diff
@@ -1,3 +1,4 @@
+
```
thijskh: Supurious whitespace is output to the browser here

tvdijen:
Two issues left in Assertion.php that I'm not sure of how to fix.

tvdijen:
Hey @dub357 !  We're about to release v5 of this module and I really really really want to include this PR since it's been open for way too long..
I've finally managed to resolve all the merge conflicts that we had, but there still are some issues.. This library went through a lot of change over the last year..
You think you can give me helping hand here to get the tests to pass?  The good thing is that, for a new major version, we don't have to keep BC-compatibility and can break stuff properly..

src/SAML2/Assertion.php
```diff
@@ -601,15 +615,15 @@ private function parseAttributeValue(DOMNode $attribute, string $attributeName):
...
 
             if ($hasNonTextChildElements) {
-                $this->attributes[$attributeName][] = $value->childNodes;
+                $this->attributes[$attributeName]->addAttributeValue(
+                    new AttributeValue($value->childNodes)
+                );
                 continue;
             }
             
-            if ($type === 'xs:integer') {
```
tvdijen: We used to allow attribute values to be an integer and then the xs:type would be properly set.. I think this PR breaks this behaviour..  It makes me wonder if the type should be just a property on the AttributeValue-class instead of an array on the Assertion-class?

dub357:
> 
> 
> Hey @dub357 ! We're about to release v5 of this module and I really really really want to include this PR since it's been open for way too long..
> I've finally managed to resolve all the merge conflicts that we had, but there still are some issues.. This library went through a lot of change over the last year..
> You think you can give me helping hand here to get the tests to pass? The good thing is that, for a new major version, we don't have to keep BC-compatibility and can break stuff properly..

It has been open a long time...quite honestly I gave up on it over a year ago for a number of reasons, but mainly because I felt like no one was making an effort to review the PR and provide feedback or approve it.  There were even other pull requests merged into the project while this one was just sitting out there, allowing it to divert further and further away from master.  But in addition to that, the whitespace nightmare that I/we originally ran into (I'm on a Windows environment) and the "strangeness" related to having all the commits to my fork post-creation of the pull request becoming part of the actual pull request itself left me with a bad taste for GitHub LOL.  

That being said, I'd still like to see this functionality make it into the library.  My original goal was to make SSP more robust with respect to attribute friendly names, name formats and xsi types...basically all the stuff that SSP couldn't do because of its rudimentary attribute handling and all of that stemmed from this library.  

I see you've been doing a lot in this fork lately.  I appreciate you pushing to get this incorporated into the next major release and I'll do what I can to help.  I can take a look at the tests, are some of them failing?  Do we need more coverage?

src/SAML2/Assertion.php
tvdijen: Similar, I think the FriendlyName should be a property of the Attribute-class...

tvdijen:
I totally feel your pain.. It's not that we don't care, but I gave up on it as well because I couldn't get rid of some of the conflicts..
Today I just went in head-first and I think it went pretty well!

What we basically need to do is get it to pass the tests (we may even break BC to accomplish this).. See https://travis-ci.org/tvdijen/saml2/jobs/624802912
I have also expressed some concerns in code-comments.. I'm especially worried about the attributeValueTypes-property on the Assertion-class because it feels misplaced.. A value type is a characteristic of AttributeValue and should probably be on that class.

Let's make this PR (finally) happen!

src/SAML2/Assertion.php
dub357: So all of the xsi:integer checks along with casts to ints currently exist in master Assertion.php.  So all of that is basically the same...not sure how this breaks anything.  
However I do notice that your fork's version of Assertion.php is very different from what I pushed to my fork years ago and tried to get into this PR.  Refer to the 'parseAttributeValue' function:
https://github.com/simplesamlphp/saml2/blob/2cba3c078149ef6f26b3610643fbf559aa46d737/src/SAML2/Assertion.php
I'm not sure what happened here, but somehow my changes didn't actually get pulled into your fork?

tvdijen: You mean those `<<<<<<< HEAD` lines? Yeah, that's what I've been working on today to get rid of.. They're all gone now?

Regarding `xsi:integer`, your original PR relied on creating AttributeValues with integers, but now that we're fully typehinted that doesn't work anymore.. Only strings are allowed, but then we need to set the type somewhere and that appears to be missing..

Maybe we should focus on the main question and the issue will be taken care of anyway;  should `xsi:sometype` be a part of the AttributeValue-class instead of keeping an array on the Assertion-class?

dub357:
> 
> 
> So looking at this now, I think perhaps I need to start over from scratch with this PR because I dont see any of the changes I was trying to introduce 2 years ago. I can take a stab at it over the holiday break. Back when I did this, I only made changes to 10 files, so it might not be to difficult. What do you think? I haven't looked at master in quite some time so I dont know how far off I would be, or even how far off your fork is from master. So, if I wanted to help with this, should I fork your branch or master?

Ok maybe nevermind...it looks like your stuff does kinda work in the same manner, its just vastly different from what I originally pushed.  Its gonna take me some time to wrap my head around the differences but I'll give it a shot

tvdijen:
I'm not sure what you mean, because all the changes in this PR originate from your [previous PR](https://github.com/simplesamlphp/saml2/pull/106).

Anyway, starting over may be the best way.. It probably already was after all the merging mess, and it makes a lot of sense after all the changes made to this library over time. I still think the changes in this PR make sense though..

@jaimeperez I imagine this PR with all the history is hard to follow, but could you give it a shot and give us direction?

src/SAML2/Assertion.php
dub357: I would say yes - absolutely.

tvdijen: OK. new starting point here:  https://github.com/simplesamlphp/saml2/tree/feature/friendlynames

dub357:
@tvdijen 
So I forked your feature-friendlynames branch the other yesterday and starting doing some playing around.  Turns out that some of the changes from my original proposed PR seem to be causing a lot of these test failures.  Mainly not allowing the assertion->setAttributes to be called without anything other than an array of Attribute objects.  From your original comment, it seems like this is because you don't want to have to worry about backwards compatibility.  Is there any reason you don't want to allow this?  Seems like its more work to have to change all the old tests rather than create new ones...

There are also a few other changes you've made that break the ability to use friendly names and name formats on a per attribute basis...especially your latest commits.   This capability was the core functionality introduced by my original PR.  The use of Attribute objects in the Assertion array was just a means to an end so to speak.   This library is the core of SSP and if one cannot control the friendly names, name formats and xsd types of produced attributes from simple attribute-based arrays in SSP config or metadata, then all this work seems like a waste.  

I also noticed that you committed/pushed a few other things to the branch after I forked.  I had even made the same change locally that you committed in [e92f28d](https://github.com/simplesamlphp/saml2/pull/122/commits/e92f28d5a0f27c8169548d0e44b6bb95f393ff8c).
I don't want to be working on a moving target and I don't want one of us to be wasting our time.    I'm tempted to simply fork master, make my changes there and push another PR, but I dont want to conflict with the existing PR you have in place - which was based on my original PR but has since deviated away quite a bit.  In that respect, I dont really know what to do at this point and some guidance would be helpful...

tvdijen:
> Mainly not allowing the assertion->setAttributes to be called without anything other than an array of Attribute objects.
I think that's exactly what we want, because it makes life much easier if we can be sure about the contents of the array.. I would say we need a `Attribute::fromArray()` to cover conversion and repair our tests

jaimeperez:
Thanks a lot for your patience @dub357! And sorry so much that this didn't progress as it should. It is mostly my fault because I have less and less time to spend on this, so I have to prioritize the stuff that's urgent.

In any case, as Tim was mentioning, I think it's a good idea to add this as a feature in a new major release, so I'll try my best to have a look into it during the Christmas holidays.

Tim, I guess we just need to change the interfaces and make them expect `Attribute` objects, or arrays of them. AFAIK, you can tell psalm you have such an array in a parameter, even though the data type will just be `array`. That would help us uncover possible bugs during the implementation. We'll need to adapt the tests as well, but I think you are already on track with that.

I'm also wondering how we could adopt this in SSP. I think it would definitely make a nice addition to 2.0, and we can't have it in place before because it definitely breaks bc. In that case, your idea to convert to/from arrays could be helpful to have an early implementation with backwards compatibility.

src/SAML2/Assertion.php
```diff
@@ -164,41 +166,13 @@ class Assertion extends SignedElement
...
-     * If the attribute is an eduPersonTargetedID, the values will be SAML2\XML\saml\NameID objects.
-     * If the attribute value has an type-definition (xsi:string or xsi:int), the values will be of that type.
-     * If the attribute value contains a nested XML structure, the values will be a DOMNodeList
-     * In all other cases the values are treated as strings
-     *
-     * **WARNING** a DOMNodeList cannot be serialized without data-loss and should be handled explicitly
-     *
-     * @var array multi-dimensional array of \DOMNodeList|\SAML2\XML\saml\NameID|string|int|array
+     * @var array array of attributes indexed by attribute name with each value an SAML2\XML\saml\Attribute 
+     * or an SAML2\XML\saml\NameID object if the attribute is an eduPersonTargetedID
```
jaimeperez: Maybe we could model this differently?:

```php
     * @var SAML2\XML\saml\Attribute[]
```

I'm also wondering about using the `NameID` here. If the goal is to allow friendly names and stuff like that, we would be missing that part if we use `NameID`s directly. In that case, I think the `NameID` should actually be the value of the `AttributeValue` object. But `AttributeValue`s use DOM objects, so I don't know exactly how to fit both...

src/SAML2/Assertion.php
```diff
@@ -1654,91 +1608,8 @@ private function addAttributeStatement(DOMElement $root): void
...
-                        $node = $document->importNode($v, true);
-                        $attributeValue->appendChild($node);
-                    }
-                } else {
-                    $value = strval($value);
-                    $attributeValue->appendChild($document->createTextNode($value));
-                }
-            }
+        foreach ($this->attributes as $name => $attributeObj) {
+            $attributeObj->toXML($attributeStatement);
```
jaimeperez: I think this is really great because it allows us to offload the responsibility to know how to marshall attributes to the attributes themselves, instead of needing code handling exceptional cases here. However, I'm not entirely sure it would work the way it is. EPTID should contain a `NameID` in its value, but I don't think that¡s the way it is implemented. Could it be an option to have a class representing this attribute in particular? It would just extend the `Attribute` class. However, we would then need a factory to create attributes.

jaimeperez: Oh, nevermind, I just saw the modifications to `AttributeValue`... 🤦‍♂

src/SAML2/Assertion.php
jaimeperez: I agree, the type of the attribute needs to be a property of the `AttributeValue` class, and its serialization routines must take care of that (e.g. calling `intval()`).

src/SAML2/Assertion.php
```diff
@@ -858,7 +830,7 @@ public function decryptNameId(XMLSecurityKey $key, array $blacklist = []): void
      */
     public function hasEncryptedAttributes(): bool
     {
-        return $this->encryptedAttributes !== [];
+        return $this->encryptedAttributes !== null && $this->encryptedAttributes !== [];
```
jaimeperez: Would this work here?:

```php
        return !empty($this->encryptedAttributes);
```

src/SAML2/Assertion.php
```diff
@@ -906,7 +878,10 @@ public function decryptAttributes(XMLSecurityKey $key, array $blacklist = []): v
             }
 
             if (!array_key_exists($name, $this->attributes)) {
-                $this->attributes[$name] = [];
+                $attr = new Attribute();
+                $attr->setName($name);
```
jaimeperez: It's a pity that the `Attribute` API uses a `DOMElement` as the parameter to the constructor. With perspective, I think we should've had a static method `fromXML()` in that class, that would return a new `Attribute` object initialized from the XML. Then the constructor would be able to receive the actual properties as arguments. That way all these code could be simplified to:

```php
$attr = new Attribute($name);
```

Most importantly, it would get rid of the inconsistency that we have today, which allows us to create an attribute without a name. This pattern makes it impossible to have objects without proper values for their mandatory properties, making the code more resilient.

Maybe it's a good time to change this behaviour as well?

src/SAML2/Assertion.php
```diff
@@ -1257,7 +1232,9 @@ public function getAttributes(): array
      */
     public function setAttributes(array $attributes): void
     {
-        $this->attributes = $attributes;
+        foreach ($attributes as $name => $value) {
+            $this->attributes[$name] = $value;
+        }
```
jaimeperez: The phpdoc here should change to reflect the actual value:

```php
     * @param Attribute[]
```

(and the same in `getAttributes()`, of course)

src/SAML2/Assertion.php
tvdijen: I think so yes

src/SAML2/Assertion.php
tvdijen: Yes, now is the time to do so!

dub357:
So I created  pull request to this branch from my fork.  It fixes all the tests but I had to make a couple tweaks to Assertion, Attribute and AttributeValue to make it work.  Take a look at let me know what you think....

src/SAML2/Assertion.php
```diff
@@ -523,96 +496,45 @@ private function parseAttributes(DOMElement $xml): void
...
-                    $hasNonTextChildElements = true;
-                    break;
-                }
-            }
-
-            $type = $value->getAttribute('xsi:type');
-            if ($type === '') {
-                $type = null;
+        if ($firstAttribute) {
+            $this->nameFormat = $nameFormat;
```
tvdijen: Why are we still keeping a NameFormat value on the Assertion-class?

src/SAML2/Assertion.php
```diff
@@ -1243,24 +1143,84 @@ public function setAuthenticatingAuthority(array $authenticatingAuthority): void
...
+     * @param \SAML2\XML\saml\Attribute[] $attributes All new attributes, as an associative array.
      * @return void
      */
     public function setAttributes(array $attributes): void
     {
-        $this->attributes = $attributes;
+        foreach ($attributes as $name => $value) {
+            if ($value instanceof Attribute) {
+                $this->attributes[$name] = $value;
+            } else {
```
tvdijen: Why do you want to keep BC-supprt for non-Attribute objects?

src/SAML2/Assertion.php
dub357: I didn’t want to rufffle the feathers too much. I’ll leave that up to you guys who “own” the codebase. I just wanted to make the tests pass. 
I’m not a huge fan of completely breaking BC but I must admit all of these changes kinda are anyway so I guess if you want to remove it, now would be the time. It really only makes sense to have when generating an assertion anyway. It just means that there will be no way to apply a single nameformat to all attributes at once but perhaps this functionality can be pushed out to SSP if it’s still needed

src/SAML2/Assertion.php
dub357: Again, I’m a fan of BC. At least for period of time while it’s made clear to folks who  are using the library that things are going to change and they should be doing things the new way. In that respect, I should have added some logging to warn about passing non-attribute objects. Again, you own this code so if you want to remove it, feel free. However I think that it would be nice to add an array typehint to ensure that the array only contains attribute objects like:

public function setAttributes(Attribute ...$attributes)

However this won’t work on PHP versions < 5.6 and I’m pretty sure we need to support 5.4

tvdijen: That would allow a variable amount of Attribute-objects passed as paramaters, not an array of Attribute-objects, or am I missing something? We don't have to support PHP5.x anymore.. v4.x of this library actually requires >=7.2

I'll discuss the BC with Jaime.. I think he's away for the holidays, so it may take a bit..

dub357: Oh oops. Well I guess there is no way I’m PHP to force an array to only contain a certain type of element unless I’m missing something. So having that code was merely a way to ensure that only Attribute objects were kept internally

PR #128, opened 2018-09-17
Fix dependency conflict with simplesamlphp/simplesamlphp

simstern:
There is a discussion in https://github.com/simplesamlphp/simplesamlphp/issues/925 about the dependency conflicts when installing simplesamlphp via composer. It suggests to use the dev-master commit directly. After merging, a new version should be tagged, so that simplesamlphp can reference it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "ext-dom": "*",
         "ext-zlib": "*",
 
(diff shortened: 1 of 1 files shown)


jaimeperez:
Thanks a lot @simstern!

Issue #130, opened 2018-10-11
SAML2\Compat\Ssp\Logger->log is not compatible with PSR-3


pradtke:
The PSR defined interface method `public function log($level, $message, array $context = array());` takes a log level. PSR-3 defines the log levels as strings

https://github.com/php-fig/log/blob/3490ba5925e6dcbe6de950c5c6b8dce9f6e96eda/Psr/Log/LogLevel.php#L10-L17

The ssp compatibility logger uses a case statement with and compares them to the SSP const log levels with are ints. As a result there are no matches and nothing gets logged.

https://github.com/simplesamlphp/saml2/blob/b3a040ec186f4a3749228b8c13642c5295fc451b/src/SAML2/Compat/Ssp/Logger.php#L126-L132

the log method should probably also handle the $context array in a similar fashion to the other methods.

PR #131, opened 2018-10-11
#130: fix psr3 log method

pradtke:
Fixes for #130

### DIFF
--- a/src/SAML2/Compat/Ssp/Logger.php
+++ b/src/SAML2/Compat/Ssp/Logger.php
@@ -3,6 +3,7 @@
 namespace SAML2\Compat\Ssp;
 
 use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
 
 class Logger implements LoggerInterface
 {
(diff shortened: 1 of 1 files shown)


src/SAML2/Compat/Ssp/Logger.php
```diff
@@ -124,29 +125,35 @@ public function debug($message, array $context = array())
...
+            case LogLevel::NOTICE:
+                $this->notice($message, $context);
                 break;
-            case \SimpleSAML\Logger::WARNING:
-                \SimpleSAML\Logger::warning($message);
+            case LogLevel::WARNING:
+                $this->warning($message, $context);
+                break;
+            default:
+                throw new \Psr\Log\InvalidArgumentException("Unrecognized log level '$level''");
```
jaimeperez: This is the only doubt that I have about this PR. Could this lead to people getting exceptions all of a sudden, when before this PR they were only missing some log messages?

pradtke: yeah, I had some concerns about that.  The PSR says

> Calling this method with a level not defined by this specification MUST throw a Psr\Log\InvalidArgumentException if the implementation does not know about the level. Users SHOULD NOT use a custom level without knowing for sure the current implementation supports it.

I guess we can say this logger "knows" about any values of $level and make the `default` case log at the debug level. Do you think we should also make the `default` case log "unrecognized level $level" as well? or just log the message and context at debug level?  I don't have a preference - I'm using this with Guzzle which only uses the PSR levels so our choice doesn't impact me.

jaimeperez: Thinking about it twice, we always use the logger via the different _`loglevel`_`()` functions, so we should never reach the default anyway. Considering the PSR says explicitly that an exception should be raised, I think your code is perfectly reasonable like that 😉

PR #132, opened 2018-10-21
Code quality (was #127)

tvdijen:
Added PHPdoc
Some scrutinizer fixes
Type hints

### DIFF
--- a/src/SAML2/Exception/UnparseableXmlException.php
+++ b/src/SAML2/Exception/UnparseableXmlException.php
@@ -14,7 +14,7 @@ final class UnparseableXmlException extends RuntimeException
     {
         $message = sprintf(
             'Unable to parse XML - "%s[%d]": "%s" in "%s" at line %d on column %d"',
-            static::$levelMap[$error->level],
+            self::$levelMap[$error->level],
             $error->code,
             $error->message,
             $error->file ?: '(string)',
--- a/src/SAML2/Message.php
+++ b/src/SAML2/Message.php
@@ -22,8 +22,7 @@ abstract class Message implements SignedElement
      *
      * @var array
      */
-    protected $extensions;
-
+    protected $extensions = array();
     /**
      * The name of the root element of the DOM tree for the message.
      *
@@ -603,8 +602,9 @@ abstract class Message implements SignedElement
     public function setExtensions($extensions)
     {
         assert(is_array($extensions) || is_null($extensions));
-
-        $this->extensions = $extensions;
+        if (is_array($extensions)) {
+            $this->extensions = $extensions;
+        }
     }
 
     /**
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "ext-dom": "*",
         "ext-zlib": "*",
 
-        "robrichards/xmlseclibs": "dev-master#93f8c07976b36b6050ab341ab5850484dc897ce0 as 3.0.2",
+        "robrichards/xmlseclibs": "^3.0",
         "psr/log": "~1.0"
     },
     "require-dev": {
--- a/src/SAML2/Assertion.php
+++ b/src/SAML2/Assertion.php
@@ -1384,8 +1384,12 @@ class Assertion implements SignedElement
 
         if (is_string($this->issuer)) {
             $issuer = Utils::addString($root, Constants::NS_SAML, 'saml:Issuer', $this->issuer);
+            $nextSibling = $issuer->nextSibling;
         } elseif ($this->issuer instanceof XML\saml\Issuer) {
             $issuer = $this->issuer->toXML($root);
+            $nextSibling = $issuer->nextSibling;
+        } else {
+            $nextSibling = null;
         }
 
         $this->addSubject($root);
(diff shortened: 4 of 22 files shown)


thijskh:
Sorry for the duplicate reviews. I think I can blame it on the Github malfunction

tvdijen:
I've tried to make the tests, but I can't do it

thijskh:
Added some tests.

tvdijen:
Done!

timmillwood:
Since this change I am getting the issue `LoginContainer::redirect($url, $data = Array) must be compatible with SAML2\Compat\AbstractContainer::redirect($url, array $data = Array)` on all sites using SAML2.

Shouldn't this BC break have gone into a new major release?

tvdijen:
Hi Tim!  I'm having a hard time to follow this one..
All I did was adding a typehint for a variable that should already be `array` or `null`..

After a redirect through some deprecated functions, we end up [here](https://github.com/simplesamlphp/simplesamlphp/blob/01865bd8d24d2f3da482fbbe6270fd65d8e61b4f/lib/SimpleSAML/Utils/HTTP.php#L963) where any value other than an array will throw an exception anyway...

So, maybe I'm just tired and missing something, but what on earth is your `LoginContainer`-class passing to this method?

thijskh:
@timmillwood What you need to do is to change your own container to add the `array` typehint for $data.

tvdijen:
I agree though that it's not nice for a bugfix-release..
It's just this one out of few cases where typehinting was forgotten...
Not sure it's worth restoring old behaviour..

timmillwood:
@thijskh right, I understand what we need to do. However in reality (partly our fault) this ended up taking down a production site as a `composer update` was done prior to deployment.

I understand typehinting may have been forgotten, but changing this in a patch release goes against semver, targeting this for the 4.x release would be awesome!

I notice this was now reverted and a new release cut, thank you so much!

src/SAML2/Exception/UnparseableXmlException.php
```diff
@@ -14,7 +14,7 @@ public function __construct(\LibXMLError $error)
     {
         $message = sprintf(
             'Unable to parse XML - "%s[%d]": "%s" in "%s" at line %d on column %d"',
-            static::$levelMap[$error->level],
+            self::$levelMap[$error->level],
```
jaimeperez: What's the reason for this change? I haven't even looked into the code, but both `self` and `static` are valid keywords with different usages, so I'm a bit puzzled that this change is happening in a code quality revision...

src/SAML2/Message.php
```diff
@@ -603,8 +602,9 @@ public function getExtensions()
     public function setExtensions($extensions)
     {
         assert(is_array($extensions) || is_null($extensions));
-
-        $this->extensions = $extensions;
+        if (is_array($extensions)) {
```
jaimeperez: I'm not sure `$this->extensions` is ever evaluated for `null` in the code (e.g.: in order to simply ignore it), but if it does, the lack of an `is_array()` check was effectively allowing us to clear the property.

thijskh:
Yes, I'm very much in favour of it but I'm worried that if SSP will not be using this new branch in the foreseeable, we will create a 'two branch' situation for a long time which is undesirable, bug prone and we also do not really have the extra time for it.

jaimeperez:
That was my initial thought as well Thijs. However, I don't think SimpleSAMLphp would be really that much affected by this. I mean, such a change could only create trouble to you if you are extending classes with signatures that changed, and those new signatures are type-hinted with types that are not allowed by your minimum PHP version supported. As in, the issue raised by @timmillwood here, but with scalar types (which are PHP 7.x). `array` and `callable`, apart from classes, are already supported by PHP 5.x.

Of course it's difficult to tell if that would be the case in SSP, but if we don't have a situation like that, it might be possible to add the saml2lib dependency as `^x.a || ^y.b`, allowing everyone to install one or the other depending on the version of PHP they are running. That would help us uncover possible bugs in SSP as well.

The argument about having to maintain two branches still holds, though, but I think it might be a bit easier to manage if the only difference between both is type declarations, as backporting changes from one to another should be trivial then.

In any case, I just wanted to stir a discussion. We can talk about this in Valencia next week, if I manage to get down there despite this terrible cold...

src/SAML2/Exception/UnparseableXmlException.php
tvdijen: Since SAML2\Exception\UnparseableXmlException is declared final, using late-static binding will have no effect.

jaimeperez: Perfect! Thanks for taking a look into it Tim 😄

tvdijen:
I've done quite some work on this today; https://github.com/tvdijen/saml2/tree/typehints
Unfortunately, I had to revert the majority of the type declarations, because for some reason PHP doesn't allow me to tighten a contract (i.e. add a type declaration that didn't exist upstream on the parent class)..

There's still quite a few tests failing that I could use some help with.. Should I create a branch + PR that we can work on further in Valencia?

PR #134, opened 2018-11-17
Fix Psalm-issues

tvdijen:
Covered by #152 
- [x] ~~Raised minimum PHP version to 7.0~~
- [x] ~~Upgraded dependencies~~
  - [x]  ~~sensiolabs/security-checker~~
  - [x]  ~~phpunit/phpunit~~
  - [x]  ~~sebastian/phpcpd~~
  - [x]  ~~mockery/mockery~~
- [x] ~~Replace PHPMD with Psalm~~
- [x] ~~Replace setExpectedException PHPdoc-annotations with actual code (after upgrading PHPunit)~~

Covered by #153:
- [x] ~~Removed all deprecated code:~~
  - [x] ~~Certificate fingerprints~~
  - [x] ~~Array-style NameId & Issuer~~
- [x] ~~Remove PSR-0 autoloader~~

Covered by #154
- [x] ~~Type declarations for scalar types~~

Covered by #155
- [x] ~~Strict types~~

Covered by #158
- [x] ~~Replace public properties with private ones~~

Remaining for this PR
- [x] PSR-2
- [x] Add more PHPdoc
- [x] Mark classes as final wherever possible
- [x] Fix remaining Psalm issues

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ php:
 matrix:
   fast_finish: true
   allow_failures:
-    - php: 7.3
     - php: hhvm
 
 before_script:
--- a/psalm.xml
+++ b/psalm.xml
@@ -3,6 +3,7 @@
   name="SimpleSAMLphp"
   useDocblockTypes="true"
   totallyTyped="false"
+  hideExternalErrors="true"
 >
   <projectFiles>
     <directory name="src/SAML2" />
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,31 +1,30 @@
 <?xml version="1.0"?>
 <psalm
-    name="SimpleSAMLphp"
-    useDocblockTypes="true"
-    totallyTyped="false"
+  name="SimpleSAMLphp"
+  useDocblockTypes="true"
+  totallyTyped="false"
 >
-    <projectFiles>
-        <directory name="src/SAML2" />
-    </projectFiles>
+  <projectFiles>
+    <directory name="src/SAML2" />
+  </projectFiles>
 
-    <issueHandlers>
-        <LessSpecificReturnType errorLevel="info" />
+  <issueHandlers>
+    <LessSpecificReturnType errorLevel="info" />
 
-        <!-- level 3 issues - slightly lazy code writing, but probably low false-negatives -->
-        <DeprecatedMethod errorLevel="info" />
+    <!-- level 3 issues - slightly lazy code writing, but probably low false-negatives -->
+    <DeprecatedMethod errorLevel="info" />
 
-        <MissingClosureReturnType errorLevel="info" />
-        <MissingReturnType errorLevel="info" />
-        <MissingPropertyType errorLevel="info" />
-        <InvalidDocblock errorLevel="info" />
-        <MisplacedRequiredParam errorLevel="info" />
+    <MissingClosureReturnType errorLevel="info" />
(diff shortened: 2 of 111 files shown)


tvdijen:
@jaimeperez @thijskh could you guys take a look at this PR as is before I continue with it?

tvdijen:
Any of you familiar with composer autoloader?
We should be able to remove the PSR-0 autoloader, but I can't get it to work;  https://travis-ci.org/simplesamlphp/saml2/jobs/458884551

tvdijen:
Coverage decrease probably has to do with the addition of this line:
https://github.com/simplesamlphp/saml2/blob/eae231e01b6d01c30ce2fe287bad317e12b6beef/tests/SAML2/HTTPPostTest.php#L66

Newer version of PHPunit would mark the test as 'Risky' without this annotation because it's not asserting anything.

tvdijen:
Scrutinizer score went from 7.68 to 7.94, number of issues went from 158 to 111

jaimeperez:
This is incredibly useful Tim!

The only thing that I miss here is that the PR was split into smaller PRs to make reviewing easier. Also, parts of it would also be useful for the older branches (like PSR-2, shorthand array syntax changes, SimpleSAMLphp API call changes or setters/getters). In particular, I think the usage of the library should be exactly the same in version 3 and 4, so that any application using it can use one version or the other depending only on the platform. That way, we can migrate SimpleSAMLphp to use the resulting backwards-compatible API, and do some actual testing over the new SAML2 library release.

I think the bare minimum we need for that is setters and getters in place for 3.x as well, but keeping the properties public. That way, we don't break backwards-compatibility, but allow at the same time to migrate to the new interface which will be the only one available in 4.x.

tvdijen:
I can split it up.. No problem...
Most of the 'generic' improvements have already been cherry-picked on master, like the shorthand arrays..
I'll start working on a PR with all the getters/setters, leaving the properties public, if that's okay with you?

jaimeperez:
That sounds great indeed, if it's not a lot of work!

tvdijen:
It IS a lot of mind-numbing work, but hey :)  Gotta do what we gotta do, right?

composer.json
```diff
@@ -15,7 +15,8 @@
         "ext-zlib": "*",
 
         "robrichards/xmlseclibs": "^3.0",
-        "psr/log": "~1.0"
+        "psr/log": "~1.0",
+        "vimeo/psalm": "^3.0"
```
jaimeperez: This was an accident, right? 😛

tvdijen: Yes, I have psalm installed locally so I can easily run tests.. I will revert this in my next push

jaimeperez:
Oh, I see, we're still using PHPUnit 4.x in this branch... 😞

jaimeperez:
Luckily, since I did that change in it's own commit, we can just revert it as soon as we have this branch rebased on top of master with the latest PHPUnit dependency...

tvdijen:
That's because I removed some of the commits that were moved to a separate PR.. Totally forgot about that :(

tvdijen:
@jaimeperez As discussed over Whatsapp, I accidentally force-pushed our work away here.. Could you force-push your local branch over this again?

tvdijen:
Great, thanks! I couldn't finish the rebase last night..
Only 123 issues left!

PR #143, opened 2018-11-29
Add getter/setter methods to replace public properties

### LINKED ISSUE #141: Public properties need to be replaced with getter/setter methods
As discussed in Valencia, the public properties have to go..
The idea is to include getter/setter functions in a 3.3 release, keeping the properties public.
Then for 4.0 we can add type hints for scalar types, remove assert()s and set the properties to private.

Work on this has started on the `gettersetter` branch
Strategy is to add the getter/setter methods and then run a `grep -R ">property;" *` and `grep -R ">property =" *` to find all assignments/lookups that need to be replaced

tvdijen:
The properties will for now remain public. They will be marked private in 4.0

Closes #141


tvdijen:
Coverage decrease is because we weren't testing the public properties before and therefore not testing the getter/setters now..
Adding tests for all of this functionality is out of scope of this PR

PR #150, opened 2018-12-18
Fix issue with loading NameID using fromArray

### LINKED ISSUE #149: NameID fromArray does not properly process SPProvidedID
A typo in SAML2\XML\saml\NameIDType::fromArray will cause the SPProvidedID value to not be loaded.

akinspe:
Fixes #149


tvdijen:
Nice catch @akinspe ! Thanks!!

PR #151, opened 2019-01-04
Provide support data: URLs for mdui:Logo

ghalse:
There are currently over 600 logos in the eduGAIN feed that are specified using RFC 2397 data: URLs. However, PHP's FILTER_VALIDATE_URL filter has no explicit support for data: URLs, and consequently makes the assumption that they must follow a `<scheme>:<host>` format. The result is that arbitrary limitations are placed on data: URLs, causing all but a very small number to fail this validation.

Ultimately this should be fixed in PHP. However, this has not happened as of 7.3.0, and even if that were to happen, we'd need to deal with the fact that the fix would be unlikely to be back ported to the earliest version supported by this library (5.4). Thus we need to have a workaround here to support data: urls in spite of PHP's bug.

cf https://github.com/php/php-src/blob/php-7.3.0/ext/filter/logical_filters.c#L587

### DIFF
--- a/src/SAML2/XML/ecp/Response.php
+++ b/src/SAML2/XML/ecp/Response.php
@@ -68,7 +68,7 @@ class Response
     {
         assert(is_string($assertionConsumerServiceURL));
         if (!filter_var($assertionConsumerServiceURL, FILTER_VALIDATE_URL)) {
-            throw new \InvalidArgumentException('Provided argument is not a valid URL.');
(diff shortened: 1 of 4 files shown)


thijskh:
Thanks! This should really have a testcase, to ensure that data: urls remain working in the future.

thijskh:
Excellent, thanks!

PR #152, opened 2019-01-17
Dependency updates

### LINKED ISSUE #147: phpmd: abandoned
PHPmd seems pretty much abandoned (last version ~ 2yrs ago)..
Given the backlog on https://github.com/phpmd/phpmd, we may want to investigate alternatives
This is especially worrying now that we're moving towards 4.0 / PHP7..

During development I've encountered an issue with phpmd, ultimately their dependencies phpdepend and symfony/dependency-injector:  https://github.com/pdepend/pdepend/issues/368

tvdijen:
These commits are a subset of #134.
Closes #147

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,6 @@
 language: php
 
 php:
-  - 5.4
-  - 5.5
-  - 5.6
   - 7.0
   - 7.1
   - 7.2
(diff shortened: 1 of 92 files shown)


tvdijen:
I'd say we merge this and start a new PR to fix all (or most) of the issues uncovered by Psalm..
Then we have a nice vantage point to start working on 4.0

tvdijen:
1. My idea was to fix the errors and then stop allowing failures for 7.3, but your plan should work as well.. I have no preference.. It's either changing the Travis-config or the Psalm-config..

2. No, these are not included yet.. My idea was to make a separate PR for that, otherwise we would end up with a huge PR.. The idea was to chop #134 into smaller bits that are easier to review..
All code changes in this PR are just to make things work with the upgraded dependencies, nothing more.

jaimeperez:
Ok, let's stick to your original plan. Both would work, and changing now would mean more work (fetch a list of all issues and whitelist them in the `psalm.xml` config), so let's avoid that.

Let's merge!

PR #153, opened 2019-01-17
Remove deprecated functionality

### LINKED ISSUE #68: Remove support for certificate validation by fingerprint
The library allows one to configure trusted certificates either in full, or with just the fingerprint. These two ways of working complicate the code significantly, while the advantage of using fingerprints seems marginal. We should therefore transition to only accept full certificates.

Discussion at https://groups.google.com/forum/#!topic/simplesamlphp-dev/MTbTGxAWniw

tvdijen:
- Issuer and NameID as objects only (remove string / array types)
- Remove PSR-0 autoloader

Closes #68


tvdijen:
0b85b64 is a rebase against master

PR #157, opened 2019-01-21
Use webmozart/assert for assertions

### LINKED ISSUE #90: Migrate assertions to something more modern

tvdijen:
Closes #90

### DIFF
--- a/src/SAML2/ArtifactResponse.php
+++ b/src/SAML2/ArtifactResponse.php
@@ -33,8 +33,6 @@ class ArtifactResponse extends StatusResponse
 
         if (!is_null($xml)) {
             $status = Utils::xpQuery($xml, './saml_protocol:Status');
-            assert(!empty($status)); /* Will have failed during StatusResponse parsing. */
-
             $status = $status[0];
 
             for ($any = $status->nextSibling; $any !== null; $any = $any->nextSibling) {
--- a/src/SAML2/Utils.php
+++ b/src/SAML2/Utils.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace SAML2;
 
-use Assert\Assertion;
+use Webmozart\Assert\Assert;
 use RobRichards\XMLSecLibs\XMLSecEnc;
 use RobRichards\XMLSecLibs\XMLSecurityDSig;
 use RobRichards\XMLSecLibs\XMLSecurityKey;
@@ -118,7 +118,7 @@ class Utils
     public static function castKey(XMLSecurityKey $key, string $algorithm, string $type = null) : XMLSecurityKey
     {
         $type = $type ?: 'public';
-        Assertion::choice($type, ["private", "public"]);
+        Assert::oneOf($type, ["private", "public"]);
 
         // do nothing if algorithm is already the type of the key
         if ($key->type === $algorithm) {
@@ -162,7 +162,7 @@ class Utils
      */
     public static function validateSignature(array $info, XMLSecurityKey $key)
     {
-        Assertion::keyExists($info, "Signature");
+        Assert::keyExists($info, "Signature");
 
         /** @var XMLSecurityDSig $objXMLSecDSig */
         $objXMLSecDSig = $info['Signature'];
--- a/src/SAML2/Utils.php
+++ b/src/SAML2/Utils.php
@@ -115,8 +115,9 @@ class Utils
      * @throws \Exception
      * @return XMLSecurityKey The new key.
      */
-    public static function castKey(XMLSecurityKey $key, string $algorithm, string $type = 'public') : XMLSecurityKey
+    public static function castKey(XMLSecurityKey $key, string $algorithm, string $type = null) : XMLSecurityKey
     {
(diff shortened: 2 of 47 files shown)


src/SAML2/ArtifactResponse.php
```diff
@@ -33,8 +33,6 @@ public function __construct(\DOMElement $xml = null)
 
         if (!is_null($xml)) {
             $status = Utils::xpQuery($xml, './saml_protocol:Status');
-            assert(!empty($status)); /* Will have failed during StatusResponse parsing. */
```
tvdijen: In case you're wondering;  this case is already being caught in the parent constructor

src/SAML2/XML/ds/KeyInfo.php
```diff
@@ -120,7 +122,7 @@ public function setInfo(array $info)
      */
     public function addInfo($info)
     {
-        assert($info instanceof Chunk || $info instanceof KeyName || $info instanceof X509Data);
+        Assertion::true($info instanceof Chunk || $info instanceof KeyName || $info instanceof X509Data);
```
jaimeperez: I was pretty sure that it was possible to check if an object was an instance of multiple alternative classes, but then checking into the API, I noticed it wasn't, and I was confused with _webmozart/assert_, which I checked back in the day. Would it make sense to switch to that other library, instead of _beberlei/assert_? The API looks almost the same, and it has the advantage that substitutions in error messages (shall we use them) can have their order specified (and therefore can be easily translated).

Sorry so much that I didn't remember about this before 😞 If it's a lot of work to switch, I'm fine with keeping this one too.

src/SAML2/XML/ds/KeyName.php
```diff
@@ -49,10 +50,10 @@ public function getName() : string
 
     /**
      * Set the value of the name-property
-     * @param string $name|null
+     * @param string $name
      * @return void
      */
-    public function setName(string $name = null)
+    public function setName(string $name)
```
jaimeperez: Is the name mandatory?

src/SAML2/XML/shibmd/Scope.php
```diff
@@ -24,7 +24,7 @@ class Scope
      *
      * @var string
      */
-    public $scope;
+    public $scope = '';
```
jaimeperez: Can the scope be an empty string?

src/SAML2/Utils.php
```diff
@@ -116,7 +117,7 @@ public static function validateElement(\DOMElement $root)
      */
     public static function castKey(XMLSecurityKey $key, string $algorithm, string $type = 'public') : XMLSecurityKey
     {
-        assert($type === "public" || $type === "private");
+        Assertion::choice($type, ["private", "public"]);
```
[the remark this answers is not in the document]
tvdijen: Great catch! I'll fix that

src/SAML2/XML/shibmd/Scope.php
jaimeperez: Consider this:

```php
            $scope = new \SAML2\XML\Shibmd\Scope();
            $string = $scope->toXML();
```

In that case, the `$scope` property will be set to an empty string, and that'll be reflected in the resulting XML. I don't know if the scope is mandatory or if it can be empty. However, there are two things we must consider:

* If the scope is not mandatory, we shouldn't be setting it to an empty string, but allow it to be `null` and skip it completely in `toXML()`. This means we should also allow clearing it, by allowing the argument to `setScope()` to be `null`.
* In that case, `getScope()` needs to be able to return `null` as well.

tvdijen: Not sure about the standards.. An empty scope doesn't seem to make much sense, but I can't find a definitive answer anywhere.. The old `assert()` in toXML would only check if it's a string or not, so a null-value would've failed.

I still think it is a good idea to set the empty string and add an assertion in the toXML-method to fail on an empty string.

jaimeperez: Given that this is shibboleth-specific, there's probably no standard about it. In any case, I agree, an empty string as the default makes things much easier for us!

PR #159, opened 2019-01-22
Push to PHP 7.2

tvdijen:
Minimum version is PHP 7.2

Upgrades PHPunit to 7.x

### DIFF
--- a/src/SAML2/AuthnRequest.php
+++ b/src/SAML2/AuthnRequest.php
@@ -40,7 +40,7 @@ class AuthnRequest extends Request
      *
      * @var string
      */
-    private $ProviderName;
+    private $ProviderName = null;
 
     /**
      * Set to true if this request is passive.
--- a/src/SAML2/AuthnRequest.php
+++ b/src/SAML2/AuthnRequest.php
@@ -410,9 +410,9 @@ class AuthnRequest extends Request
     /**
      * Retrieve the value of the ProviderName attribute.
      *
-     * @return string The ProviderName attribute.
+     * @return string|null The ProviderName attribute.
      */
-    public function getProviderName() : string
+    public function getProviderName() : ?string
     {
         return $this->ProviderName;
     }
--- a/src/SAML2/AuthnRequest.php
+++ b/src/SAML2/AuthnRequest.php
@@ -4,12 +4,15 @@ declare(strict_types=1);
 
 namespace SAML2;
 
+use DOMDocument;
+use DOMElement;
 use RobRichards\XMLSecLibs\XMLSecEnc;
 use RobRichards\XMLSecLibs\XMLSecurityKey;
+use Webmozart\Assert\Assert;
 
+use SAML2\XML\saml\NameID;
 use SAML2\XML\saml\SubjectConfirmation;
 use SAML2\Exception\InvalidArgumentException;
-use Webmozart\Assert\Assert;
 
 /**
  * Class for SAML 2 authentication request messages.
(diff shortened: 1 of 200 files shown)


tvdijen:
HHVM is now definitely broken..

src/SAML2/AuthnRequest.php
```diff
@@ -35,9 +38,9 @@ class AuthnRequest extends Request
     /**
      * Optional ProviderID attribute
      *
-     * @var string|null
+     * @var string
      */
-    private $ProviderName = null;
+    private $ProviderName;
```
jaimeperez: As far as I see in the specs, the `ProviderName` is optional, meaning this should default to `null`. Otherwise, we would always be adding a `ProviderName` even if we just have an empty string as the default.

src/SAML2/XML/md/KeyDescriptor.php
```diff
@@ -30,9 +31,9 @@ class KeyDescriptor
     /**
      * The KeyInfo for this key.
      *
-     * @var \SAML2\XML\ds\KeyInfo|null
+     * @var \SAML2\XML\ds\KeyInfo
      */
-    private $KeyInfo = null;
+    private $KeyInfo;
```
jaimeperez: This one can still be `null`, and should therefore default to that. Even though the standard mandates it as required, we can have a situation where we create a `KeyDescriptor` object without passing any XML to it, and therefore if we call `getKeyInfo()` immediately after that, the property would still be `null`. I don't know if that would trigger an error, since the `getKeyInfo()` method has now a return type specified, and we would be returning `null` instead, but in any case, we should reflect the actual case.

tvdijen: That should trigger a TypeError, yes.. And toXML would also fail because of the assert in place

jaimeperez: The exception due to the assert in `toXML()` would be normal behaviour, but the `TypeError` wouldn't. It's kind of a grey zone, because `null` would only be returned when the API is misused, but still I think we should actually return the value reflecting the object has not been set, instead of raising an error.

tvdijen:
Psalm really doesn't like those assertions... And neither do I..
What a mess to get the tests to pass again

jaimeperez:
It makes sense, after all. It's not checking the source code for the _webmozart/assert_ library, so it doesn't have any possible way to know that after the assertion, the variable is guaranteed to have a value.

In any case, I'd say psalm is serving its purpose very well... 😄

tvdijen:
Can we specifically include the webmozart source to prevent these kind of issues in the future?

tvdijen:
Changes to the webmozart library are minimal, looking at their history.. I'd say we go for it.
Worst case scenario, we have to suppress (or cast to info) specific errors on a per-file basis.

Given the nature of the xmlseclibs library, we may want to include that one too. Also not a library that has too many changes.

jaimeperez:
xmlseclibs is failing hard nowadays. I submitted a [PR](https://github.com/robrichards/xmlseclibs/pull/188) to fix the issues, but would need that merged (and psalm integrated as part of the build process as well) before we can include it in our own tests.

tvdijen:
I know.. He left you a review comment there

jaimeperez:
I need to add a new commit there, yes...

tvdijen:
How do you feel about writing our own [Stub-files](https://github.com/vimeo/psalm/blob/master/docs/configuration.md#project-settings) for XMLSecLibs and Webmozart? As an alternative to including them in the Psalm-config

tvdijen:
OK, stubs are nice for php-libs, not for the webmozart asserts.
Also, including the vendor/webmozart/assert dir to the project didn't have the expected result

PR #169, opened 2019-03-12
Replace assertions with Webmozart

### LINKED ISSUE #168: Settings variables in assert()
https://github.com/simplesamlphp/saml2/blob/94d61b2f5eeb7aa96784c87f9d58b8c7992fbc0f/src/SAML2/XML/mdui/DiscoHints.php#L156-L164

This section doesn't work when PHP's zend.assertions option is < 1, which is the case if you're using the php.ini-production file which ships with PHP 7.2 (or anything derived from it).  Because PHP doesn't execute the assert(), the local variables are never set, so the if condition can never be true.

This affects SimpleSAMLphp 1.17.1: even if I set DiscoHints in metadata/saml20-idp-hosted.php, they simply aren't included in the metadata that's output.

tvdijen:
Fixes and closes #168

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 5.4
   - 5.5
   - 5.6
   - 7.0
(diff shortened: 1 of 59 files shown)


tvdijen:
Gah, I pull it against master instead of 3.x

PR #170, opened 2019-03-13
Webmozart assertions

### LINKED ISSUE #168: Settings variables in assert()

tvdijen:
Fixes and closes #168

### DIFF
--- a/src/SAML2/AuthnRequest.php
+++ b/src/SAML2/AuthnRequest.php
@@ -4,8 +4,10 @@ namespace SAML2;
 
 use RobRichards\XMLSecLibs\XMLSecEnc;
 use RobRichards\XMLSecLibs\XMLSecurityKey;
+use SAML2\XML\saml\NameID;
 use SAML2\XML\saml\SubjectConfirmation;
 use SAML2\Exception\InvalidArgumentException;
+use Webmozart\Assert\Assert;
 
 /**
  * Class for SAML 2 authentication request messages.
@@ -206,7 +208,7 @@ class AuthnRequest extends Request
             /* The NameID element is encrypted. */
             $this->encryptedNameId = $nameId;
         } else {
-            $this->nameId = new XML\saml\NameID($nameId);
+            $this->nameId = new NameID($nameId);
         }
 
         $subjectConfirmation = Utils::xpQuery($subject, './saml_assertion:SubjectConfirmation');
@@ -389,7 +391,7 @@ class AuthnRequest extends Request
      */
     public function setForceAuthn($forceAuthn)
     {
-        assert(is_bool($forceAuthn));
+        Assert::boolean($forceAuthn);
 
         $this->forceAuthn = $forceAuthn;
     }
@@ -414,7 +416,7 @@ class AuthnRequest extends Request
      */
     public function setProviderName($ProviderName)
     {
-        assert(is_string($ProviderName));
+        Assert::string($ProviderName);
 
         $this->ProviderName = $ProviderName;
     }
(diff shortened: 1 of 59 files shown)


src/SAML2/AuthnRequest.php
```diff
@@ -660,10 +662,8 @@ public function getRequestedAuthnContext()
      * @param array $requestedAuthnContext The RequestedAuthnContext.
      * @return void
      */
-    public function setRequestedAuthnContext($requestedAuthnContext)
+    public function setRequestedAuthnContext(array $requestedAuthnContext)
```
jaimeperez: This is not something we can add in the 3.x branch, right? Type hinting with scalar types is not available in PHP 5.x.

src/SAML2/Message.php
```diff
@@ -226,10 +228,8 @@ private function validateSignature(\DOMElement $xml)
      * @param mixed    $data     The data that should be included as the first parameter to the function
      * @return void
      */
-    public function addValidator($function, $data)
+    public function addValidator(callable $function, $data)
```
jaimeperez: Same here, I think callable is not available for type hinting in PHP 5.x.

src/SAML2/SignedElementHelper.php
```diff
@@ -77,10 +77,8 @@ protected function __construct(\DOMElement $xml = null)
      * @param mixed    $data     The data that should be included as the first parameter to the function.
      * @return void
      */
-    public function addValidator($function, $data)
+    public function addValidator(callable $function, $data)
```
jaimeperez: ⬆️

src/SAML2/XML/md/EntityDescriptor.php
```diff
@@ -450,16 +451,16 @@ public function addAdditionalMetadataLocation(AdditionalMetadataLocation $additi
...
-        assert(is_array($this->getContactPerson()));
-        assert(is_array($this->getAdditionalMetadataLocation()));
+        Assert::string($this->getEntityID());
+        Assert::nullOrString($this->getID());
+        Assert::nullOrInteger($this->getValidUntil());
+        Assert::nullOrString($this->getCacheDuration());
+        Assert::isArray($this->getExtensions());
+        Assert::isArray($this->getRoleDescriptor());
+        Assert::nullOrIsInstanceOf($this->getAffiliationDescriptor());
+        Assert::nullOrIsInstance($this->getOrganization());
```
jaimeperez: Does `nullOrIsInstance()` exist, or is this a typo?

src/SAML2/AuthnRequest.php
tvdijen: http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration

Since 5.1

src/SAML2/Message.php
tvdijen: Since 5.4
http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration

src/SAML2/SignedElementHelper.php
tvdijen: Since 5.4
http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration

src/SAML2/XML/md/EntityDescriptor.php
tvdijen: That's a typo, thanks!

PR #182, opened 2019-10-02
Do not decrypt already decrypted assertions

### LINKED ISSUE #181: Trying to process response and failing due to double decryption
Trying to process response fails with **`UnencryptedAssertionFoundException`** exception saying that `The assertion should be encrypted, but it was not`.

Looking it seems like it's being decrypted twice and an exception is thrown for that reason.

First time it's decrypted in `processAssertions`
```php
    private function processAssertions(Response $response) : ArrayCollection
    {
        $assertions = $response->getAssertions();
        if (empty($assertions)) {
            throw new NoAssertionsFoundException('No assertions found in response from IdP.');
        }

        $decryptedAssertions = $this->assertionProcessor->decryptAssertions(
            new ArrayCollection($assertions)
        );
       <...>
```
Later it is trying to decrypt it again and fails:
```php
    public function process($assertion) : Assertion
    {
        $assertion = $this->decryptAssertion($assertion);
        <...>
```

AndriusKuc:
Fix an issue of trying to decrypt already decrypted assertion.
This Fixes #181

### DIFF
--- a/src/SAML2/Assertion/Processor.php
+++ b/src/SAML2/Assertion/Processor.php
@@ -152,12 +152,12 @@ class Processor
      */
     private function decryptAssertion($assertion): Assertion
     {
-        if ($this->decrypter->isEncryptionRequired() && $assertion instanceof Assertion) {
-            throw new UnencryptedAssertionFoundException('The assertion should be encrypted, but it was not');
-        }
-
         if ($assertion instanceof Assertion) {
             return $assertion;
+        }
+
+        if ($this->decrypter->isEncryptionRequired() && $assertion instanceof Assertion) {
+            throw new UnencryptedAssertionFoundException('The assertion should be encrypted, but it was not');
         }
 
         return $this->decrypter->decrypt($assertion);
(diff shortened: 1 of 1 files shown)


tvdijen:
The first commit makes full sense and should fix your issue.. The second only adds unnecessary complexity... Why would you do that?

AndriusKuc:
Looking back at changes there was a comment - https://github.com/simplesamlphp/saml2/pull/120#issuecomment-385907102 that suggested in case somebody is uging  assertion processor as a stand-alone component we wouldn't want to break how it works.

tadasauciunas:
> The first commit makes full sense and should fix your issue.. The second only adds unnecessary complexity... Why would you do that?

Code that uses uses decryptAssertion might expect UnencryptedAssertionFoundException if a SAML2\Assertion object is passed into the method instead of SAML2\EncryptedAssertion. That would break backwards-compatibility. Andrius' second commit fixes that issue, and calls decryptAssertion during procession only if it's actually necessary.

chris1312:
@tvdijen are you expecting any other contributions to this issue?
We would like to move on and finish upgrading SAML version on our project :)

tvdijen:
We're discussing if we're going to break BC for the sake of a proper fix, since this is a relatively new major release that hasn't really been adopted by other projects yet.. We would just remove the call to decryptAssertion completely..

PR #185, opened 2019-11-14
Minor improvements

tvdijen:
Raise an exception when mandatory fields are missing instead of returning null


src/SAML2/Assertion.php
```diff
@@ -1575,7 +1575,7 @@ private function addConditions(DOMElement $root): void
     private function addAuthnStatement(DOMElement $root): void
     {
         if (
-            $this->authnInstant === null
+            !isset($this->authnInstant)
```
jaimeperez: And these two as well.

src/SAML2/LogoutRequest.php
```diff
@@ -276,7 +276,7 @@ public function setSessionIndex(string $sessionIndex = null): void
      */
     public function toUnsignedXML(): DOMElement
     {
-        if ($this->encryptedNameId === null && $this->nameId === null) {
+        if ($this->encryptedNameId === null && !isset($this->nameId)) {
```
jaimeperez: In this file as well. Was psalm complaining about the comparison because the phpdoc said it couldn't be `null`?

src/SAML2/Assertion.php
```diff
@@ -418,7 +418,7 @@ private function parseAuthnStatement(DOMElement $xml): void
         /** @var \DOMElement[] $authnStatements */
         $authnStatements = Utils::xpQuery($xml, './saml_assertion:AuthnStatement');
         if (empty($authnStatements)) {
-            $this->authnInstant = null;
+            unset($this->authnInstant);
```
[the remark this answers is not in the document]
tvdijen: I may be mistaken, but I think Psalm will complain about assigning null to a string-property..

src/SAML2/LogoutRequest.php
tvdijen: Yes, I think it complains about a RedundantCondition when explicitly testing for null

src/SAML2/Assertion.php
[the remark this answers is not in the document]
jaimeperez: Makes sense. I was expecting that. In a way, I think the key here is that the phpdoc was correct. The attribute can have a type or be `null`, and we need to account for that anyway.

src/SAML2/SubjectQuery.php
```diff
@@ -77,21 +78,23 @@ private function parseSubject(\DOMElement $xml): void
     /**
      * Retrieve the NameId of the subject in the query.
      *
-     * @return \SAML2\XML\saml\NameID|null The name identifier of the assertion.
+     * @return \SAML2\XML\saml\NameID The name identifier of the assertion.
      */
-    public function getNameId(): ?NameID
+    public function getNameId(): NameID
     {
+        Assert::notEmpty($this->nameId);
```
jaimeperez: I was playing a bit before with some examples, and PHP definitely throws an exception when you try to return a type that doesn't match the signature of the function. However, I don't know how psalm would deal with that case. It's probably a good idea to be explicit here anyway.

src/SAML2/XML/md/AffiliationDescriptor.php
```diff
@@ -251,7 +253,7 @@ public function addKeyDescriptor(KeyDescriptor $keyDescriptor): void
      */
     public function toXML(DOMElement $parent): DOMElement
     {
-        Assert::notEmpty($this->AffiliateMember);
```
jaimeperez: Actual bug?

src/SAML2/Assertion.php
```diff
@@ -1487,6 +1486,8 @@ public function toXML(\DOMNode $parentElement = null): DOMElement
         $root->setAttribute('Version', '2.0');
         $root->setAttribute('IssueInstant', gmdate('Y-m-d\TH:i:s\Z', $this->issueInstant));
 
+        Assert::notEmpty($this->issuer, 'Cannot convert Assertion to XML without an Issuer set.');
```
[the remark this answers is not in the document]
tvdijen: Taken care of

src/SAML2/XML/md/AffiliationDescriptor.php
tvdijen: Yup, this was a bug

tvdijen: I'll backport it to 4.x and mark this as resolved

PR #189, opened 2019-12-19
Add samlp:NameIDPolicy class

tvdijen:
At this point we're still processing NameIDPolicy as arrays instead of objects


tvdijen:
Discussion [here](https://github.com/simplesamlphp/saml2/pull/122#discussion_r358698253) got me wondering whether I should go over the library again to see if I could improve the PHPdocs..

At this point we're still processing NameIDPolicy as arrays instead of objects

tvdijen:
Tests fail right now.. do we want a fromArray-method?

tests/SAML2/AuthnRequestTest.php
```diff
@@ -676,8 +674,10 @@ public function testSettingNameIDPolicyToIncorrectTypeSPNameQualifier(): void
         $request->setDestination('https://tiqr.example.org/idp/profile/saml2/Redirect/SSO');
 
         // SPNameQualifier must be a string
-        $nameIdPolicy = ["SPNameQualifier" => true];
-        $this->expectException(\InvalidArgumentException::class, 'Invalid Argument type: "string" expected');
+        $nameIdPolicy = new NameIDPolicy();
+        $this->expectException(\TypeError::class);
```
tvdijen: The only question I ask myself about this PR is whether it is still desirable to test for invalid argument types now that we have fully typehinted the codebase..

jaimeperez: I guess psalm would uncover these errors, so tests can then focus on uncovering business logic errors rather than programming ones...

tvdijen: Exactly! Great! I'll remove those tests..  I don't think they help for coverage anyway, since the method called isn't run at all..

tvdijen:
I've taken care of all comments..  Now that the constructor is rewritten, I wonder.. Should we make the setter-methods private and only allow creation through the constructor?

The reason I ask is that right now we have a possibility to end up with an object with all of the properties set to `null`. which will generate `<samlp:NameIDPolicy />`.. While that's not necessarily a problem, we probably want to leave it out completely.. Right now the only way I can prevent an empty tag is to raise an exception in toXML.. Private settings would allow me to assert this 'all null' case in the constructor + fromXML

jaimeperez:
Wouldn't it be possible to achieve the same goal making the type hint in the setter _non-nullable_?

In any case, it depends on the standard. The `toXML()` method should never generate an XML document that doesn't comply with the standard.

tvdijen:
Making it non-nullable would take away the possibility to unset it..
I'm really not sure what the best solution is here

jaimeperez:
I was assuming the standard mandates at least some of the properties. Those properties that are mandatory should not be nullable, and therefore I don't see a need to unset them. But I might be looking at the wrong place...

tvdijen:
https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf
Paragraph 3.4.1.1

thijskh:
According to the spec all attributes are optional. If the user of the library choses to set none of them, but asks for an element, it will be empty. I see no problem in that.

MKodde:
Nice work @tvdijen I'd like to contribute my opinion to some aspects that have not yet been discussed. Let me start by saying that I think merging this PR was a good call.

> My suggestion is to make the setters private and force every object creation through the constructor, at the cost of not being able to change values at a later point in time..

Thinking about this. When would you ever want to change these values? There might be some valid use cases (that I can not come up with but others might?). 

In the case you must modify an immutable object, you can simply keep it immutable by returning a new object with the new value(s). 

> Right now the only way to prevent this is to kill execution with an assertion in toXML, which is pretty harsh.. Protecting the constructor to have at least one of the optional settings set is probably a better solution. However, the setter-methods can still lead us into a broken/undesirable situation..

Personally I would add guards (assertions) to the constructor, ensuring no invalid instances of the object can be constructed. If an invalid set of parameters is provided, throwing an exception is the go-to solution. You could throw a 'we don't want empty objects exception' when all values are empty. Alternatively, you could speculate that the factory logic, tasked with calling the NameIdPolicy constructor, could prevent creating an empty instance.

Thijs' solution fits well too, adding an empty instance wouldn't hurt.

> If the user of the library choses to set none of them, but asks for an element, it will be empty. I see no problem in that.

If the action is user-initiated yes, when the library itself is creating the document, you might not want to add the empty object.

tvdijen:
Thanks for your comment @MKodde!

> Thinking about this. When would you ever want to change these values?
Like in my example when you want to add an AttributeValue to an Attribute, but that could be handled by bullet-proof adder-methods.. I guess in 99% of the cases objects can be perfectly immutable, hence my suggestion to bullet-proof them and make the setters private..

>If the action is user-initiated yes, when the library itself is creating the document, you might not want to add the empty object.
Unfortunately we cannot differentiate the two cases in the toXML-methods.. Even in the user-initiated case I wouldn't want to add the empty tag, because it's basically just token bloat..

I've merged this and some other PR's for now, because #192/#193/#194 are quite a big improvements already.. We can always decide to change this later if we can agree on a way to go..

PR #190, opened 2019-12-22
add samlp:Status[Detail|Code|Message] classes

tvdijen:
We used to treat them as arrays.. StatusDetail wasn't support before

### DIFF
--- a/tests/SAML2/StatusResponseTest.php
+++ b/tests/SAML2/StatusResponseTest.php
@@ -4,9 +4,13 @@ declare(strict_types=1);
 
 namespace SAML2;
 
+use SAML2\DOMDocumentFactory;
 use SAML2\Response;
 use SAML2\Utils;
-use SAML2\DOMDocumentFactory;
+use SAML2\XML\samlp\Status;
+use SAML2\XML\samlp\StatusCode;
+use SAML2\XML\samlp\StatusMessage;
+use SAML2\XML\samlp\StatusDetail;
 
 /**
  * Class \SAML2\StatusResponseTest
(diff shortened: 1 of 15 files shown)


tests/SAML2/StatusResponseTest.php
```diff
@@ -18,12 +22,21 @@ class StatusResponseTest extends \PHPUnit\Framework\TestCase
...
+        $subcode = new StatusCode();
+        $subcode->setValue('OurSubStatusCode');
+        $code = new StatusCode();
+        $code->setValue('OurStatusCode');
+        $code->setSubCodes([$subcode]);
+        $status->setStatusCode($code);
+
+        $message = new StatusMessage();
+        $message->setMessage('OurMessageText');
+        $status->setStatusMessage($message);
```
jaimeperez: With an API inverted to work the way we are discussing, this could be simplified to:

```php
        $status = new Status(
            new StatusCode(
                'OurStatusCode',
                [new StatusCode('OurSubStatusCode')]
            ),
            new StatusMessage('OurMessageText')
        ),
```

Is that easier to read?

tests/SAML2/XML/samlp/StatusCodeTest.php
```diff
@@ -0,0 +1,66 @@
...
+
+        $document = DOMDocumentFactory::fromString('<root />');
+        /** @psalm-var \DOMElement $document->firstChild */
+        $statusCodeElement = $statusCode->toXML($document->firstChild);
+
+        $this->assertEquals(Constants::STATUS_RESPONDER, $statusCodeElement->getAttribute('Value'));
+
+        /** @psalm-var \DOMElement[] $statusCodeElements */
+        $statusCodeElements = Utils::xpQuery($statusCodeElement, './saml_protocol:StatusCode');
+        $this->assertEquals(Constants::STATUS_REQUEST_DENIED, $statusCodeElements[0]->getAttribute('Value'));
```
jaimeperez: One idea that I was considering is to make all these classes inherit from a basic class that allows getting a string with the XML. Something like an interface with an abstract `toXML()`, and an implementation of `__toString()` similar to this: 

```php
    public function __toString()
    {
        $xml = $this->toXML();
        return $xml->ownerDocument->saveXML($xml);
    }
```

That would ease tests, because we then just need to compare the resulting XML with an actual string, eliminating the need to process XML manually ourselves.

tvdijen: Yeah that would be great! We can add an abstract fromXML to the interface as well

PR #191, opened 2019-12-24
Add classes for RequestedAuthnContext structure

tvdijen:
Adds classes + unit tests for:
RequestedAuthnContext
AuthnContextClassRef
AuthnContextDeclRef

Implements these classes into the codebase

### DIFF
--- a/src/SAML2/XML/samlp/RequestedAuthnContext.php
+++ b/src/SAML2/XML/samlp/RequestedAuthnContext.php
@@ -60,17 +60,15 @@ class RequestedAuthnContext
     public function setRequestedAuthnContexts(array $requestedAuthnContexts): void
     {
         Assert::minCount($requestedAuthnContexts, 1);
-        $types = array_unique(array_map('gettype', $requestedAuthnContexts));
+        Assert::allIsInstanceOfAny($requestedAuthnContexts, [AuthnContextClassRef::class, AuthnContextDeclRef::class]);
 
         if ($requestedAuthnContexts[0] instanceof AuthnContextClassRef) {
-            Assert::allIsInstanceOf($requestedAuthnContexts, AuthnContextClassRef::class);
+            Assert::allIsInstanceOf($requestedAuthnContexts, AuthnContextClassRef::class, 'You need either AuthnContextClassRef or AuthnContextDeclRef, not both.');
         } elseif ($requestedAuthnContexts[0] instanceof AuthnContextDeclRef) {
-            Assert::allIsInstanceOf($requestedAuthnContexts, AuthnContextDeclRef::class);
+            Assert::allIsInstanceOf($requestedAuthnContexts, AuthnContextDeclRef::class, 'You need either AuthnContextClassRef or AuthnContextDeclRef, not both.');
         } else {
-            throw new \InvalidArgumentException('Invalid object types provided.');
+            throw new \InvalidArgumentException('You need either AuthnContextClassRef or AuthnContextDeclRef.');
         }
-
-        Assert::count($types, 1, 'You need either AuthnContextClassRef or AuthnContextDeclRef, not both');
 
         $this->requestedAuthnContexts = $requestedAuthnContexts;
     }
@@ -116,7 +114,7 @@ class RequestedAuthnContext
         Assert::oneOf(
             [],
             [$authnContextClassRef, $authnContextDeclRef],
-            'You need either AuthnContextClassRef or AuthnContextDeclRef, not both'
+            'You need either AuthnContextClassRef or AuthnContextDeclRef, not both.'
         );
 
         $requestedAuthnContexts = [];
--- a/src/SAML2/XML/samlp/RequestedAuthnContext.php
+++ b/src/SAML2/XML/samlp/RequestedAuthnContext.php
@@ -1,0 +1,169 @@
+<?php
+
+declare(strict_types=1);
+
+namespace SAML2\XML\samlp;
+
+use DOMElement;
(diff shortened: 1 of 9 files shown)


src/SAML2/XML/samlp/RequestedAuthnContext.php
```diff
@@ -0,0 +1,142 @@
...
+            1,
+            'You need either AuthnContextClassRef or AuthnContextDeclRef, not both'
+        );
+
+        foreach ($this->requestedAuthnContexts as $context) {
+            $e->appendChild($e->ownerDocument->importNode($context->toXML(), true));
+        }
+
+        if (isset($this->Comparison)) {
+            Assert::oneOf($this->Comparison, ['exact', 'minimum', 'maximum', 'better']);
```
jaimeperez: This assertion is probably better in `setComparison()`. That way, if it fails, it fails early and in the place where you did something wrong (assigning an erroneous value), rather than when converting to XML.

tests/SAML2/XML/samlp/RequestedAuthnContextTest.php
```diff
@@ -0,0 +1,102 @@
...
+</samlp:RequestedAuthnContext>
+XML
+        );
+
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('You need either AuthnContextClassRef or AuthnContextDeclRef, not both');
+
+        /** @psalm-var \DOMElement $document->firstChild */
+        $requestedAuthnContext = new RequestedAuthnContext($document->firstChild);
+    }
```
jaimeperez: We should probably have a test for the opposite scenario, trying to marshall into XML when you have added both `AuthnContextClassRef` and `AuthnContextDeclrRef`.

src/SAML2/XML/samlp/RequestedAuthnContext.php
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+
+    /**
+     * Set the value of the Comparison-property
+     *
+     * @param string|null $comparison
+     * @return void
+     */
+    public function setComparison(?string $comparison): void
```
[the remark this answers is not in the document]
tvdijen: Yes, that's the nullable typehint... What you are suggesting would make it optional..
My intent here is to require a value, that may be `null` to be able to unset the current value...

src/SAML2/XML/samlp/RequestedAuthnContext.php
```diff
@@ -0,0 +1,142 @@
...
+     *
+     * @param \DOMElement $parent The element we should append this RequestedAuthnContext to.
+     * @return \DOMElement
+     */
+    public function toXML(DOMElement $parent): DOMElement
+    {
+        $e = $parent->ownerDocument->createElementNS(Constants::NS_SAMLP, 'samlp:RequestedAuthnContext');
+        $parent->appendChild($e);
+
+        $types = array_unique(array_map('gettype', $this->requestedAuthnContexts));
```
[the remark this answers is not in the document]
tvdijen: Thanks! Took me a while to have this properly asserted :P

src/SAML2/XML/samlp/RequestedAuthnContext.php
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * Set the value of the requestedAuthnContexts-property
+     *
+     * @param (\SAML2\XML\saml\AuthnContextClassRef|\SAML2\XML\saml\AuthnContextDeclRef)[] $requestedAuthnContexts
+     * @return void
+     */
+    public function setRequestedAuthnContexts(array $requestedAuthnContexts): void
+    {
+        $this->requestedAuthnContexts = $requestedAuthnContexts;
```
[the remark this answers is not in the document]
tvdijen: Probably yes.. I'll add an Assert both here and in to toXML-method

src/SAML2/XML/samlp/RequestedAuthnContext.php
tvdijen: If we do that, we have to use the setter-method in the constructor as well.. That way we can never import XML with an unspecified value

tests/SAML2/XML/samlp/RequestedAuthnContextTest.php
tvdijen: Consider it done!

tvdijen:
In general I agree with the constructor + fromXML method, but let's not do that in this PR for just one class

src/SAML2/XML/samlp/RequestedAuthnContext.php
tvdijen: So to add to this, we basically need to decide whether we do our sanity/compliancy checks on setting the values or when generating the output XML..  Previously we tested the output to make sure it's compliant, but personally I'm in favour of catching issues on the input-side..

src/SAML2/XML/samlp/RequestedAuthnContext.php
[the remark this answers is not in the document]
jaimeperez: I see! I thought both were equivalent, but if not, then you can just disregard my comment 😅

src/SAML2/XML/samlp/RequestedAuthnContext.php
jaimeperez: Agreed, I think the sooner, the better. I also thought about calling the setter in the constructor. If _unspecified_ is legal (which I don't know), then the constructor could only call the setter if there was a value specified.

PR #197, opened 2020-01-07
Add classes for AuthnContext structure

tvdijen:
Adds classes + unit tests for:
AuthnContext
AuthnContextDecl
AuthenticatingAuthority

Implements these classes into the codebase

### DIFF
--- a/src/SAML2/XML/AbstractConvertable.php
+++ b/src/SAML2/XML/AbstractConvertable.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace SAML2\XML;
 
 use DOMElement;
+use SAML2\DOMDocumentFactory;
+use Serializable;
 
 /**
  * Abstract class to be implemented by all the classes in this namespace
@@ -13,7 +15,7 @@ use DOMElement;
  * @package SimpleSAMLphp
  */
 
-abstract class AbstractConvertable
+abstract class AbstractConvertable implements Serializable
 {
     /**
      * Output the class as an XML-formatted string
@@ -24,6 +26,37 @@ abstract class AbstractConvertable
     {
         $xml = $this->toXML();
         return $xml->ownerDocument->saveXML($xml);
+    }
+
+
+    /**
+     * Serialize this XML chunk
+     *
+     * @return string The serialized chunk.
+     */
+    public function serialize(): string
+    {
+        return $this->toXML()->ownerDocument->saveXML();
+    }
+
+
+    /**
+     * Un-serialize this XML chunk.
+     *
+     * @param string $serialized The serialized chunk.
+     * @return void
+     *
+     * Type hint not possible due to upstream method signature
+     */
+    public function unserialize($serialized): void
+    {
+        $doc = DOMDocumentFactory::fromString($serialized);
+        $obj = static::fromXML($doc->documentElement);
+
+        // For this to work, the properties have to be protected
+        foreach (get_object_vars($obj) as $property => $value) {
+            $this->{$property} = $value;
+        }
     }
 
 
--- a/src/SAML2/XML/saml/AuthenticatingAuthority.php
+++ b/src/SAML2/XML/saml/AuthenticatingAuthority.php
@@ -1,0 +1,97 @@
+<?php
+
+declare(strict_types=1);
+
+namespace SAML2\XML\saml;
+
+use DOMElement;
+use SAML2\Constants;
+use SAML2\DOMDocumentFactory;
+use SAML2\Utils;
+use Webmozart\Assert\Assert;
+
+/**
+ * Class representing SAML2 AuthenticatingAuthority
+ *
+ * @author Tim van Dijen, <tvdijen@gmail.com>
+ * @package SimpleSAMLphp
+ */
+class AuthenticatingAuthority extends \SAML2\XML\AbstractConvertable
+{
+    /** @var string */
+    private $authority;
+
+
+    /**
+     * Initialize an AuthicatingAuthority.
+     *
+     * @param string $authority
+     */
+    public function __construct(string $authority)
+    {
+        $this->setAuthority($authority);
+    }
+
+
+    /**
+     * Collect the value of the authority-property
+     *
+     * @return string
+     *
+     * @throws \InvalidArgumentException if assertions are false
+     */
+    public function getAuthority(): string
+    {
+        return $this->authority;
+    }
+
+
+    /**
+     * Set the value of the authority-property
+     *
+     * @param string $name
+     * @return void
+     */
+    private function setAuthority(string $authority): void
+    {
+        $this->authority = trim($authority);
+    }
+
+
+    /**
+     * Convert XML into a AuthenticatingAuthority
+     *
+     * @param \DOMElement $xml The XML element we should load
+     * @return \SAML2\XML\saml\AuthenticatingAuthority
+     */
+    public static function fromXML(DOMElement $xml): object
+    {
+        Assert::same($xml->tagName, 'saml:AuthenticatingAuthority');
+        Assert::same($xml->namespaceURI, Constants::NS_SAML);
+
+        return new self($xml->textContent);
+    }
+
+
+    /**
+     * Convert this AuthenticatingAuthority to XML.
+     *
+     * @param \DOMElement|null $parent The element we should append this AuthnContextClassRef to.
+     * @return \DOMElement
+     */
+    public function toXML(DOMElement $parent = null): DOMElement
+    {
+        if ($parent === null) {
+            $doc = DOMDocumentFactory::create();
+            $e = $doc->createElementNS(Constants::NS_SAML, 'saml:AuthenticatingAuthority');
+            $doc->appendChild($e);
+        } else {
+            $e = $parent->ownerDocument->createElementNS(Constants::NS_SAML, 'saml:AuthenticatingAuthority');
+            $parent->appendChild($e);
+        }
+
+        $e->textContent = $this->authority;
+
+        return $e;
+    }
+}
--- a/src/SAML2/XML/saml/AuthenticatingAuthority.php
+++ b/src/SAML2/XML/saml/AuthenticatingAuthority.php
@@ -66,7 +66,7 @@ class AuthenticatingAuthority extends \SAML2\XML\AbstractConvertable
      */
     public static function fromXML(DOMElement $xml): object
     {
-        Assert::same($xml->tagName, 'saml:AuthenticatingAuthority');
+        Assert::same($xml->localName, 'AuthenticatingAuthority');
         Assert::same($xml->namespaceURI, Constants::NS_SAML);
 
         return new self($xml->textContent);
--- a/src/SAML2/XML/saml/AuthenticatingAuthority.php
+++ b/src/SAML2/XML/saml/AuthenticatingAuthority.php
@@ -16,10 +16,10 @@ use Webmozart\Assert\Assert;
  * @author Tim van Dijen, <tvdijen@gmail.com>
  * @package SimpleSAMLphp
  */
-class AuthenticatingAuthority extends \SAML2\XML\AbstractConvertable
+final class AuthenticatingAuthority extends \SAML2\XML\AbstractConvertable
 {
     /** @var string */
-    private $authority;
+    protected $authority;
 
 
     /**
--- a/src/SAML2/XML/saml/AuthnContext.php
+++ b/src/SAML2/XML/saml/AuthnContext.php
@@ -20,19 +20,19 @@ use Webmozart\Assert\Assert;
  * @author Tim van Dijen, <tvdijen@gmail.com>
  * @package SimpleSAMLphp
  */
-class AuthnContext extends \SAML2\XML\AbstractConvertable
+final class AuthnContext extends \SAML2\XML\AbstractConvertable
 {
     /** @var \SAML2\XML\saml\AuthnContextClassRef|null */
-    private $authnContextClassRef = null;
+    protected $authnContextClassRef = null;
 
     /** @var \SAML2\XML\saml\AuthnContextDeclRef|null */
-    private $authnContextDeclRef = null;
+    protected $authnContextDeclRef = null;
 
     /** @var \SAML2\XML\saml\AuthnContextDecl|null */
-    private $authnContextDecl = null;
+    protected $authnContextDecl = null;
 
     /** @var \SAML2\XML\saml\AuthenticatingAuthority[]|null */
-    private $authenticatingAuthorities = null;
+    protected $authenticatingAuthorities = null;
 
 
     /**
--- a/src/SAML2/XML/saml/AuthnContext.php
+++ b/src/SAML2/XML/saml/AuthnContext.php
@@ -1,0 +1,240 @@
+<?php
+
+declare(strict_types=1);
+
+namespace SAML2\XML\saml;
+
+use DOMElement;
+use SAML2\Constants;
+use SAML2\DOMDocumentFactory;
+use SAML2\Utils;
+use SAML2\XML\saml\AuthenticatingAuthority;
+use SAML2\XML\saml\AuthnContextClassRef;
+use SAML2\XML\saml\AuthnContextDecl;
+use SAML2\XML\saml\AuthnContextDeclRef;
+use Webmozart\Assert\Assert;
+
+/**
+ * Class representing SAML2 AuthnContext
+ *
+ * @author Tim van Dijen, <tvdijen@gmail.com>
+ * @package SimpleSAMLphp
+ */
+class AuthnContext extends \SAML2\XML\AbstractConvertable
+{
+    /** @var \SAML2\XML\saml\AuthnContextClassRef|null */
+    private $authnContextClassRef = null;
+
+    /** @var \SAML2\XML\saml\AuthnContextDeclRef|null */
+    private $authnContextDeclRef = null;
+
+    /** @var \SAML2\XML\saml\AuthnContextDecl|null */
+    private $authnContextDecl = null;
+
+    /** @var \SAML2\XML\saml\AuthenticatingAuthority[]|null */
+    private $authenticatingAuthorities = null;
+
+
+    /**
+     * Initialize an AuthnContext.
+     *
+     * @param \SAML2\XML\saml\AuthnContextClassRef|null $authnContextClassRef
+     * @param \SAML2\XML\saml\AuthnContextDecl|null $authnContextDecl
+     * @param \SAML2\XML\saml\AuthnContextDeclRef|null $authnContextDeclRef
+     * @param \SAML2\XML\saml\AuthenticatingAuthority[]|null $authenticatingAuthorities
+     */
+    public function __construct(
+        ?AuthnContextClassRef $authnContextClassRef,
+        ?AuthnContextDecl $authnContextDecl,
+        ?AuthnContextDeclRef $authnContextDeclRef,
+        ?array $authenticatingAuthorities
+    ) {
+        if (!is_null($authnContextClassRef)) {
+            Assert::oneOf(
+                null,
+                [$authnContextDecl, $authnContextDeclRef],
+                'Can only have one of AuthnContextDecl/AuthnContextDeclRef'
+            );
+        } else {
+            Assert::false(
+                is_null($authnContextDecl) && is_null($authnContextDeclRef),
+                'You need either an AuthnContextDecl or an AuthnContextDeclRef'
+            );
+        }
+
+        $this->setAuthnContextClassRef($authnContextClassRef);
+        $this->setAuthnContextDecl($authnContextDecl);
+        $this->setAuthnContextDeclRef($authnContextDeclRef);
+        $this->setAuthenticatingAuthorities($authenticatingAuthorities);
+    }
+
+
+    /**
+     * Collect the value of the authnContextClassRef-property
+     *
+     * @return \SAML2\XML\saml\AuthnContextClassRef|null
+     */
+    public function getAuthnContextClassRef(): ?AuthnContextClassRef
+    {
+        return $this->authnContextClassRef;
+    }
+
+
+    /**
+     * Set the value of the authnContextClassRef-property
+     *
+     * @param \SAML2\XML\saml\AuthnContextClassRef|null $authnContextClassRef
+     * @return void
+     */
+    private function setAuthnContextClassRef(?AuthnContextClassRef $authnContextClassRef): void
+    {
+        $this->authnContextClassRef = $authnContextClassRef;
+    }
+
+
+    /**
+     * Collect the value of the authnContextDeclRef-property
+     *
+     * @return \SAML2\XML\saml\AuthnContextDeclRef|null
+     */
+    public function getAuthnContextDeclRef(): ?AuthnContextDeclRef
+    {
+        return $this->authnContextDeclRef;
+    }
+
+
+    /**
+     * Set the value of the authnContextDeclRef-property
+     *
+     * @param \SAML2\XML\saml\AuthnContextDeclRef|null $authnContextDeclRef
+     * @return void
+     */
+    private function setAuthnContextDeclRef(?AuthnContextDeclRef $authnContextDeclRef): void
+    {
+        $this->authnContextDeclRef = $authnContextDeclRef;
+    }
+
+
+    /**
+     * Collect the value of the authnContextDecl-property
+     *
+     * @return \SAML2\XML\saml\AuthnContextDecl|null
+     */
+    public function getAuthnContextDecl(): ?AuthnContextDecl
+    {
+        return $this->authnContextDecl;
+    }
+
+
+    /**
+     * Set the value of the authnContextDecl-property
+     *
+     * @param \SAML2\XML\saml\AuthnContextDecl|null $authnContextDecl
+     * @return void
+     */
+    private function setAuthnContextDecl(?AuthnContextDecl $authnContextDecl): void
+    {
+        $this->authnContextDecl = $authnContextDecl;
+    }
+
+
+    /**
+     * Collect the value of the authenticatingAuthorities-property
+     *
+     * @return \SAML2\XML\saml\AuthenticatingAuthority[]|null
+     */
+    public function getAuthticatingAuthorities(): ?array
+    {
+        return $this->authenticatingAuthorities;
+    }
+
+
+    /**
+     * Set the value of the authenticatingAuthorities-property
+     *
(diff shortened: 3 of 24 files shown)


tvdijen:
Just a quick sanity check will suffice

src/SAML2/XML/saml/AuthnContextDecl.php
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Class representing SAML2 AuthnContextDecl
+ *
+ * @author Tim van Dijen, <tvdijen@gmail.com>
+ * @package SimpleSAMLphp
+ */
+class AuthnContextDecl extends \SAML2\XML\AbstractConvertable
+{
+    /** @var \DOMElement|null */
```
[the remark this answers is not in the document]
jaimeperez: Actually, I was wondering... would it make sense to use `\SAML2\XML\Chunk` in here, instead of `DOMElement` directly?

src/SAML2/XML/saml/AuthnContext.php
```diff
@@ -0,0 +1,266 @@
...
+     * @param \SAML2\XML\saml\AuthenticatingAuthority[]|null $authenticatingAuthorities
+     */
+    public function __construct(
+        ?AuthnContextClassRef $authnContextClassRef,
+        ?AuthnContextDecl $authnContextDecl,
+        ?AuthnContextDeclRef $authnContextDeclRef,
+        ?array $authenticatingAuthorities
+    ) {
+        if (!is_null($authnContextClassRef)) {
+            Assert::oneOf(
```
jaimeperez: Does `oneOf` allow the case of both being `null`? Because in this case, the minimum occurs is set to zero, so both could be skipped in fact.

src/SAML2/XML/saml/AuthnContext.php
```diff
@@ -0,0 +1,266 @@
...
+        } else {
+            Assert::false(
+                is_null($authnContextDecl) && is_null($authnContextDeclRef),
+                'You need either an AuthnContextDecl or an AuthnContextDeclRef'
+            );
+        }
+
+        $this->setAuthnContextClassRef($authnContextClassRef);
+
+        if (!is_null($authnContextDecl)) {
```
jaimeperez: Given that the setters have nullable parameters, we can skip these `is_null()` checks and just call the setters.

src/SAML2/XML/saml/AuthnContext.php
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+
+    /**
+     * Set the value of the authnContextClassRef-property
+     *
+     * @param \SAML2\XML\saml\AuthnContextClassRef|null $authnContextClassRef
+     * @return void
+     */
+    public function setAuthnContextClassRef(?AuthnContextClassRef $authnContextClassRef): void
```
jaimeperez: I would also mark this as protected at least. The simpler the interface, the better, so if we set everything in the constructor, there's no point in having public setters, IMHO.

src/SAML2/XML/saml/AuthnContext.php
```diff
@@ -0,0 +1,266 @@
...
+
+    /**
+     * Set the value of the authenticatingAuthorities-property
+     *
+     * @param \SAML2\XML\saml\AuthenticatingAuthority[]|null $authenticatingAuthorities
+     * @return void
+     */
+    public function setAuthenticatingAuthorities(?array $authenticatingAuthorities): void
+    {
+        $this->authenticatingAuthorities = $authenticatingAuthorities;
```
jaimeperez: Here I think we should check the elements and verify that all of them are `SAML2\XML\saml\AuthenticatingAuthority` objects.

src/SAML2/XML/saml/AuthnContext.php
```diff
@@ -0,0 +1,266 @@
...
+                [],
+                [$authnContextDecl, $authnContextDeclRef],
+                'Can only have one of AuthnContextDecl/AuthnContextDeclRef'
+            );
+        } else {
+            Assert::false(
+                empty($authnContextDecl) && empty($authnContextDeclRef),
+                'You need either an AuthnContextDecl or an AuthnContextDeclRef'
+            );
+        }
```
jaimeperez: We also have this exact check in the constructor. Can we skip it here, considering we're just passing the processed values to it later in the method?

src/SAML2/XML/saml/AuthnContext.php
```diff
@@ -0,0 +1,266 @@
...
+
+        /** @var \DOMElement[] $authenticatingAuthorities */
+        $authenticatingAuthorities = Utils::xpQuery($xml, './saml_assertion:AuthenticatingAuthority');
+
+        return new self(
+            empty($authnContextClassRef) ? null : AuthnContextClassRef::fromXML($authnContextClassRef[0]),
+            empty($authnContextDecl) ? null : AuthnContextDecl::fromXML($authnContextDecl[0]),
+            empty($authnContextDeclRef) ? null : AuthnContextDeclRef::fromXML($authnContextDeclRef[0]),
+            array_map([AuthenticatingAuthority::class, 'fromXML'], $authenticatingAuthorities) ?: null
```
jaimeperez: 👏

src/SAML2/XML/saml/AuthnContext.php
```diff
@@ -0,0 +1,266 @@
...
+                null,
+                [$this->authnContextDecl, $this->authnContextDeclRef],
+                'Can only have one of AuthnContextDecl/AuthnContextDeclRef'
+            );
+        } else {
+            Assert::false(
+                empty($this->authnContextDecl) && empty($this->authnContextDeclRef),
+                'You need either an AuthnContextDecl or an AuthnContextDeclRef'
+            );
+        }
```
jaimeperez: If the setters are protected (or private), then every possible path to create an object will lead to this check in the constructor. Then the check becomes redundant here.

tests/SAML2/XML/saml/AuthenticatingAuthorityTest.php
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Class \SAML2\XML\saml\AuthenticatingAuthorityTest
+ */
+class AuthenticatingAuthorityTest extends \PHPUnit\Framework\TestCase
+{
+    /**
+     * @return void
+     */
+    public function testMarshalling(): void
```
jaimeperez: With the new `AbstractConvertable` class, we can now simplify tests like this by casting the object to a string and comparing with a string representing the XML document.

src/SAML2/XML/saml/AuthenticatingAuthority.php
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Set the value of the authority-property
+     *
+     * @param string $name
+     * @return void
+     */
+    public function setAuthority(string $authority): void
+    {
+        $authority = trim($authority);
+        Assert::stringNotEmpty($authority);
```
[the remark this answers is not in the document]
tvdijen: That's a good catch.. I'll remove the assertion and double-check the already merged PRs

src/SAML2/XML/saml/AuthenticatingAuthority.php
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+
+    /**
+     * Set the value of the authority-property
+     *
+     * @param string $name
+     * @return void
+     */
+    public function setAuthority(string $authority): void
```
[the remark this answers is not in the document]
tvdijen: That's exactly what I was suggesting in #189.. We should do that everywhere and make the objects immutable, but that's a design choice I can't make on my own..

src/SAML2/XML/saml/AuthnContext.php
tvdijen: OK, if that's what we agree upon, then I'm going to change that for the already merged PR's as well

src/SAML2/XML/saml/AuthenticatingAuthority.php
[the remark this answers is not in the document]
jaimeperez: I think the approach to make objects immutable would help us reduce complexity and have easier to use interfaces, so let's go ahead 👍

src/SAML2/XML/saml/AuthnContextDecl.php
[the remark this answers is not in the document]
tvdijen: I think so, but where would we do the conversion? In the constructor? In the calling code?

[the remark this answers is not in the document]
jaimeperez: Intuitively, I'd say in the calling code. Then, if your code supports a specific declaration, you can always model that extending `Chunk`, and still pass it around here...

src/SAML2/XML/saml/AuthnContext.php
tvdijen: Yes, that is allowed. I've double checked by adding an `Assert::oneOf(null, [null, null], 'This is wrong');` and it didn't trigger the error message..

jaimeperez: Perfect!

src/SAML2/XML/saml/AuthnContext.php
tvdijen: Yeah, that was quite interesting :P

src/SAML2/XML/saml/AuthnContextDecl.php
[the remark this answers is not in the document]
tvdijen: I've pushed the requested changes.. Can you give it another look?

src/SAML2/XML/saml/AuthnContextDecl.php
```diff
@@ -68,7 +67,7 @@ public function setDecl(DOMElement $decl): void
      */
     public static function fromXML(DOMElement $xml): object
     {
-        return new self($xml);
+        return new self(new Chunk($xml));
```
jaimeperez: I'm not sure about this one. What you would pass to the `fromXML()` methods are the elements themselves, right? In this case, the `<saml:AuthnContextDecl>` element. The chunk in this case would then be the contents inside `$xml`.

src/SAML2/XML/saml/AuthenticatingAuthority.php
```diff
@@ -69,6 +66,9 @@ public function setAuthority(string $authority): void
      */
     public static function fromXML(DOMElement $xml): object
     {
+        Assert::same($xml->tagName, 'saml:AuthenticatingAuthority');
+        Assert::same($xml->namespaceURI, Constants::NS_SAML);
```
jaimeperez: Great! These checks should also be applied to the other elements, I think.

src/SAML2/XML/saml/AuthenticatingAuthority.php
```diff
@@ -92,7 +90,7 @@ public function toXML(DOMElement $parent = null): DOMElement
             $parent->appendChild($e);
         }
 
-        $e->textContent = $this->authority;
+        $e->textContent = $this->getAuthority();
```
jaimeperez: Why this change?

src/SAML2/XML/saml/AuthenticatingAuthority.php
tvdijen: I will do that

src/SAML2/XML/saml/AuthnContextDecl.php
tvdijen: Hmm, then how is it possible for the tests to pass?

src/SAML2/XML/saml/AuthenticatingAuthority.php
tvdijen: My bad... I dunno what I was thinking, but it's undone

src/SAML2/XML/saml/AuthenticatingAuthority.php
tvdijen: I've added a commit for AuthnContext.. Can't do AuthnContextDecl because that one can contains virtually anything

src/SAML2/XML/saml/AuthnContextDecl.php
tvdijen: Hmm, I see I do pass the elements themselves..

tvdijen: I guess you've discovered a bug.. I've commit a fix

tvdijen:
If you could give it another go please? All comments have been addressed and I added a commit that fixes the already merged PR's

src/SAML2/XML/saml/AuthnContextClassRef.php
```diff
@@ -69,6 +65,9 @@ public function setClassRef(string $classRef): void
      */
     public static function fromXML(DOMElement $xml): object
     {
+        Assert::same($xml->tagName, 'saml:AuthnContextClassRef');
```
jaimeperez: This should check `$xml->localName` instead. The namespace prefix can be anything, the standard does not mandate it to be `saml`.

tvdijen: That goes for all of them I guess?

jaimeperez: Yep, we should always use `localName` when comparing.

src/SAML2/XML/saml/AuthnContextDecl.php
```diff
@@ -0,0 +1,99 @@
...
+     * @param \DOMElement $xml The XML element we should load
+     * @return \SAML2\XML\saml\AuthnContextDecl
+     */
+    public static function fromXML(DOMElement $xml): object
+    {
+        Assert::same($xml->tagName, 'saml:AuthnContextDecl');
+        Assert::same($xml->namespaceURI, Constants::NS_SAML);
+
+        /** @psalm-var \DOMElement $xml->childNodes[1] */
+        return new self(new Chunk($xml->childNodes[1]));
```
jaimeperez: I was already thinking about a problem with this before seeing your implementation, and now I see it confirmed 😞. The contents of this element can be anything, meaning these two would be valid:

```xml
<saml:AuthnContextDecl>
  <foo>bar</foo>
</saml:AuthnContextDecl>
```

```xml
<saml:AuthnContextDecl>
  <foo>1</foo>
  <bar>2</bar>
</saml:AuthnContextDecl>
```

Both are valid, but the current implementation would only work with the former. The latter would yield an object that contains only the `<foo>` element. Even worse, this would also be valid:

```xml
<saml:AuthnContextDecl>
  foo
  <bar>2</bar>
</saml:AuthnContextDecl>
```

I don't see an easy way out, but I draw some conclusions:

1. We can't assume the number of nodes children of the element to be only 1.
2. We can't use `SAML2\XML\Chunk` with the contents, because that's intended for XML elements, but the contents can be of any type, and in an arbitrary number.

I think probably the best solution would be to recursively import all the child nodes into a new `DOMDocument` and append them there. The `DOMDocument` would then be the property we save, and the class needs to implement `\Serializable` the same way `SAML2\XML\Chunk` does it.

tvdijen: Now I see why nobody ever implemented AuthnContextDecl and StatusDetail :rofl:

src/SAML2/XML/samlp/Status.php
```diff
@@ -66,7 +67,7 @@ public function getStatusCode(): StatusCode
      *
      * @throws \InvalidArgumentException if assertions are false
      */
-    public function setStatusCode(StatusCode $statusCode): void
+    private function setStatusCode(StatusCode $statusCode): void
```
jaimeperez: I'm not completely convinced we should declare all setters as private. In some cases, I think it would be desirable to extend an existing class to implement improved functionality, and then access to those methods would be needed. Do we already have any classes under `SAML2\XML` that extend others?

src/SAML2/XML/samlp/StatusDetail.php
```diff
@@ -64,7 +65,11 @@ public function setDetail(?\DOMElement $detail): void
      */
     public static function fromXML(DOMElement $xml): object
     {
-        return new self($xml);
+        Assert::same($xml->tagName, 'samlp:StatusDetail');
+        Assert::same($xml->namespaceURI, Constants::NS_SAMLP);
+
+        /** @psalm-var \DOMElement $xml->childNodes[1] */
+        return new self(new Chunk($xml->childNodes[1]));
```
jaimeperez: Having a look at the core document of the standard, here we also have a problem with the contents. The good news, in this case, is that the contents are supposed to be XML elements indeed, so we could use an array of `SAML2\XML\Chunk` objects.

tests/SAML2/XML/saml/AuthenticatingAuthorityTest.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        $authority = new AuthenticatingAuthority('https://sp.example.com/SAML2');
+
+        $this->assertEquals(
+            strval($authority),
+            '<saml:AuthenticatingAuthority xmlns:saml="' . Constants::NS_SAML
+                . '">https://sp.example.com/SAML2</saml:AuthenticatingAuthority>'
+        );
```
jaimeperez: A minor detail here: the parameters to `assertEquals()` are the expected value and the actual value, in that order, so we should probably invert them here to avoid confusions in case of an error.

src/SAML2/XML/saml/AuthnContextDecl.php
jaimeperez: 😆

tvdijen: https://stackoverflow.com/questions/5389076/difference-similarities-between-xsdany-and-xsdanytype
I think the 3rd case is out of the question, so that would leave us with an array of Chunks here as well

src/SAML2/XML/samlp/Status.php
tvdijen: We extend from SignedElementHelper, RoleDescriptor, SSODescriptorType, EndpointType, NameIDType and Attribute..

I can only imagine doing so if you want to support some almost-but-not-really-SAML spec.. Strictly speaking the classes should be final, but I guess that's pretty harsh..

jaimeperez: I actually like the approach to make classes final better than making setters private. In the end, the standard has its own extension points, where you should add your functionality, whatever that shall be. We should follow the same principle, and the rest of the classes representing elements defined by SAML2 should probably be final as you are saying, unless we have a reason not to make them so (e.g. we're extending them somewhere, but in that case we're talking about types, rather than actual elements, so for those it wouldn't make sense to mark them as final.

I'm wondering if we should use traits instead of an abstract class that we are extending...

src/SAML2/XML/saml/AuthnContextDecl.php
tvdijen: Should be done.. Please take a look

src/SAML2/XML/samlp/Status.php
tvdijen: If we make them public again, we have to add a lot of protecting assertions to toXML again to avoid illegal situations.. I thought we had agreed on the immutable objects..
I'd say final+private and/or non-final+protected..

src/SAML2/XML/samlp/StatusDetail.php
tvdijen: I read this a but too late, so I took the same approach as AuthnContextDecl.. A later commit reverts this and uses Chunks..

src/SAML2/XML/AbstractConvertable.php
```diff
@@ -27,6 +29,36 @@ public function __toString(): string
...
 
 
+    /**
+     * Serialize this XML chunk
+     *
+     * @return string The serialized chunk.
+     */
+    public function serialize(): string
+    {
+        return serialize($this->toXML()->ownerDocument->saveXML());
```
jaimeperez: There's no need to call `serialize()` here. The `serialize()` method of the `Serializable` interface needs to return a string that you can later recover, but that doesn't mean it needs to follow any particular format. Returning the XML document as a string should suffice, as we can recover it later from that representation.

src/SAML2/XML/AbstractConvertable.php
```diff
@@ -27,6 +29,36 @@ public function __toString(): string
...
+     * Un-serialize this XML chunk.
+     *
+     * @param string $serialized The serialized chunk.
+     * @return void
+     *
+     * Type hint not possible due to upstream method signature
+     */
+    public function unserialize($serialized): void
+    {
+        $doc = DOMDocumentFactory::fromString(unserialize($serialized));
```
jaimeperez: ... and the same here. No need to call `unserialize()`.

src/SAML2/XML/saml/AuthenticatingAuthority.php
```diff
@@ -0,0 +1,97 @@
...
+    public function toXML(DOMElement $parent = null): DOMElement
+    {
+        if ($parent === null) {
+            $doc = DOMDocumentFactory::create();
+            $e = $doc->createElementNS(Constants::NS_SAML, 'saml:AuthenticatingAuthority');
+            $doc->appendChild($e);
+        } else {
+            $e = $parent->ownerDocument->createElementNS(Constants::NS_SAML, 'saml:AuthenticatingAuthority');
+            $parent->appendChild($e);
+        }
```
jaimeperez: This block is a pattern that repeats a lot. It might be an idea to move it to a generic method inside `AbstractConvertable` receiving the namespace and the name of the element we want to create, plus the parent. Then all these blocks of identical code can be substituted with a single call to that method.

src/SAML2/XML/saml/AuthnContextDecl.php
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Un-serialize this XML chunk.
+     *
+     * @param string $serialized The serialized chunk.
+     * @return void
+     *
+     * Type hint not possible due to upstream method signature
+     */
+    public function unserialize($serialized): void
```
jaimeperez: This shouldn't be necessary, I believe. The parent calling `fromXML()` should be enough to recreate the object, right?

src/SAML2/XML/AbstractConvertable.php
tvdijen: Done

src/SAML2/XML/AbstractConvertable.php
tvdijen: And done

src/SAML2/XML/samlp/Status.php
jaimeperez: No, no, I didn't mean making them public, but protected. But of course that would only make sense for non-final classes. Also, provided that the setters are protected, I'm fine with the properties being private, since any class potentially extending can still get and set those values.

src/SAML2/XML/samlp/StatusCode.php
```diff
@@ -97,6 +97,9 @@ public function setSubCodes(?array $subCodes): void
      */
     public static function fromXML(DOMElement $xml): object
     {
+        Assert::same($xml->localName, 'StatusCode');
+        Assert::same($xml->namespaceURI, Constants::NS_SAMLP);
+
```
jaimeperez: I can't add a comment on the proper line, so I'll add it here. Line 147:

```php
        if (!empty($this->subCodes)) {
            foreach ($this->subCodes as $subCode) {
                $subCode->toXML($e);
            }
        }
```

The outer if can be removed I think, since if `$this->subCodes` is empty, the loop will not iterate.

src/SAML2/XML/samlp/StatusDetail.php
```diff
@@ -18,41 +20,73 @@
...
+        $this->details = $details;
+    }
+
+
+    /**
+     * Serialize this XML chunk
+     *
+     * @return string The serialized chunk.
+     */
+    public function serialize(): string
```
jaimeperez: `serialize()` and `unserialize()` shouldn't be necessary here either. `fromXML()` should handle reconstruction of the object.

src/SAML2/XML/samlp/StatusDetail.php
```diff
@@ -64,7 +98,11 @@ public function setDetail(?\DOMElement $detail): void
      */
     public static function fromXML(DOMElement $xml): object
     {
-        return new self($xml);
+        Assert::same($xml->localName, 'StatusDetail');
+        Assert::same($xml->namespaceURI, Constants::NS_SAMLP);
+
+        /** @psalm-var \DOMElement $xml->childNodes[1] */
+        return new self([new Chunk($xml->childNodes[1])]);
```
jaimeperez: This is a bug. We should iterate over the `DOMNodeList` in `$xml->childNodes` and create a new `SAML2\XML\Chunk` for each `DOMElement`, then pass an array of all those chunks to the constructor.

src/SAML2/XML/samlp/StatusCode.php
tvdijen: It's optional, so it could be null..

src/SAML2/XML/samlp/StatusDetail.php
```diff
@@ -85,8 +123,10 @@ public function toXML(DOMElement $parent = null): DOMElement
             $parent->appendChild($e);
         }
 
-        if (!empty($this->detail)) {
-            $e->appendChild($e->ownerDocument->importNode($this->detail, true));
+        if (!empty($this->details)) {
```
jaimeperez: Oh, I can actually comment on this one 😅

Same case as before, the `if` is redundant in this case, as `foreach` tests the condition.

src/SAML2/XML/samlp/StatusDetail.php
tvdijen: Removed

tests/SAML2/XML/saml/AuthenticatingAuthorityTest.php
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Class \SAML2\XML\saml\AuthenticatingAuthorityTest
+ */
+class AuthenticatingAuthorityTest extends \PHPUnit\Framework\TestCase
+{
+    /**
+     * @return void
+     */
+    public function testMarshalling(): void
```
jaimeperez: These tests look wonderful now, so simple and easy to read! ❤️

src/SAML2/XML/samlp/StatusCode.php
jaimeperez: My bad, didn't notice it could be null!

src/SAML2/XML/samlp/StatusDetail.php
tvdijen: Has been resolved

PR #198, opened 2020-01-10
Migrate XML/alg to new interface

tvdijen:


src/SAML2/XML/alg/DigestMethod.php
```diff
@@ -65,27 +56,47 @@ public function getAlgorithm(): string
...
+     * @param \DOMElement|null $parent The element we should append to.
      * @return \DOMElement
      *
      * @throws \InvalidArgumentException if assertions are false
      */
-    public function toXML(DOMElement $parent): DOMElement
+    public function toXML(DOMElement $parent = null): DOMElement
     {
-        Assert::notEmpty($this->Algorithm, 'Cannot convert DigestMethod to XML without an Algorithm set.');
+        if ($parent === null) {
```
jaimeperez: Maybe it's a good idea to have a very simple PR first with this code in a method inside `AbstractConvertable`, and then use it in these PRs instead of having to copy/paste the code?

tvdijen: Sure, let's do that right after this one.. Unless you've spotted issues?

jaimeperez: Haven't finished yet, but not so far...

tests/SAML2/XML/alg/SigningMethodTest.php
```diff
@@ -34,25 +33,10 @@ public function testMarshalling(): void
         );
         $this->assertCount(1, $signingMethodElements);
         $signingMethodElement = $signingMethodElements[0];
+
         $this->assertEquals('http://exampleAlgorithm', $signingMethodElement->getAttribute('Algorithm'));
-        $this->assertFalse($signingMethodElement->hasAttribute('MinKeySize'));
+        $this->assertEquals('1024', $signingMethodElement->getAttribute('MinKeySize'));
```
jaimeperez: Should we be a bit more exhaustive with the tests? As in, testing all possible scenarios of missing (or not) attributes...

tvdijen: We used to do all attributes set and optional attributes missing.. I reduced it to a mixed scenario, but I can easily revert that..

src/SAML2/XML/alg/DigestMethod.php
tvdijen: The method was added, but it depends on #199 to be merged

PR #199, opened 2020-01-12
Create intermediate abstracts per namespace

tvdijen:


src/SAML2/XML/AbstractXMLElement.php
```diff
@@ -60,6 +67,29 @@ public function unserialize($serialized): void
...
 
+    /**
+     * Create a document structure for this element
+     *
+     * @param \DOMElement|null $parent The element we should append to.
+     * @return \DOMElement
+     */
+    public function instantiateParentElement(DOMElement $parent = null): DOMElement
+    {
+        $qualifiedName = join('', array_slice(explode('\\', get_class()), -1));
```
jaimeperez: I think this is wrong. Calling `get_class()` without params assumes the current class, which in this case is `SAML2\XML\AbstractXMLElement`. We should either be using `get_called_class()` or `static::class`, since that will use late static binding and therefore point to the right class.

jaimeperez: Just to demonstrate it:

```php
php > class A {
php {     public static function m() {
php {         echo get_class();
php {     }
php { }
php > class B extends A {}
php > echo A::m();
A
php > echo B::m();
A
```

src/SAML2/XML/saml/AbstractSamlElement.php
```diff
@@ -0,0 +1,23 @@
...
+ * @author Tim van Dijen, <tvdijen@gmail.com>
+ * @package SimpleSAMLphp
+ */
+abstract class AbstractSamlElement extends AbstractXMLElement
+{
+    /** @var string */
+    public const NS = Constants::NS_SAML;
+
+    /** @var string */
+    public const NS_PREFIX = 'saml';
```
jaimeperez: These are not strictly necessary since those are the defaults defined in `AbstractConvertable`.

src/SAML2/XML/saml/AuthenticatingAuthority.php
```diff
@@ -16,7 +16,7 @@
  * @author Tim van Dijen, <tvdijen@gmail.com>
  * @package SimpleSAMLphp
  */
-final class AuthenticatingAuthority extends \SAML2\XML\AbstractConvertable
+final class AuthenticatingAuthority extends \SAML2\XML\AbstractXMLElement
```
jaimeperez: This one (and the rest in the `src/SAML2/XML/saml/` directory) should extend from `SAML2\XML\saml\AbstractSamlElement`.

src/SAML2/XML/samlp/NameIDPolicy.php
```diff
@@ -15,7 +15,7 @@
  * @author Tim van Dijen, <tvdijen@gmail.com>
  * @package SimpleSAMLphp
  */
-final class NameIDPolicy extends \SAML2\XML\AbstractConvertable
+final class NameIDPolicy extends \SAML2\XML\AbstractXMLElement
```
jaimeperez: These ones are incorrect as well, they should extend from `SAML2\XML\samlp\AbstractSamlpElement`.

src/SAML2/XML/saml/AbstractSamlElement.php
tvdijen: I'd rather remove them from `AbstractXMLElement` then..

src/SAML2/XML/AbstractXMLElement.php
tvdijen: Did I fix it wrong? Or would my solution work as well?

jaimeperez: Both are equivalent, yep. I just didn't see your fix while I was taking a look at the PR 😅

PR #200, opened 2020-01-12
Migrate XML/ds to new interface

tvdijen:


src/SAML2/XML/ds/X509Certificate.php
```diff
@@ -60,24 +51,35 @@ public function getCertificate(): string
...
      * @throws \InvalidArgumentException if assertions are false
      */
-    public function toXML(DOMElement $parent): DOMElement
+    public function toXML(DOMElement $parent = null): DOMElement
     {
-        Assert::notEmpty($this->certificate);
-
-        return Utils::addString($parent, XMLSecurityDSig::XMLDSIGNS, 'ds:X509Certificate', $this->certificate);
+//        $e = $this->instantiateParentElement($parent);
+        return Utils::addString($parent, self::NS, 'ds:X509Certificate', $this->certificate);
```
tvdijen: This is one outstanding issue that occurs twice in this PR..
Our interface requires `toXML` to accept null. so I had to change it..
Passing `$e` to `Utils::addString` didn't work because it leads to `<ds:X509Certificate><ds:X509Certificate>somecert</ds:X509Certificate></ds:X509Certificate>`..

tvdijen:
And through the magic 70% coverage barrier! :-D

src/SAML2/XML/ds/KeyInfo.php
```diff
@@ -31,45 +30,19 @@ class KeyInfo
...
 
     /**
      * Initialize a KeyInfo element.
      *
-     * @param \DOMElement|null $xml The XML element we should load.
+     * @param (\SAML2\XML\Chunk|\SAML2\XML\ds\KeyName|\SAML2\XML\ds\X509Data)[] $info
+     * @param string|null $Id
      */
-    public function __construct(DOMElement $xml = null)
+    public function __construct(array $info, $Id = null)
```
jaimeperez: Shouldn't we specify the `string` type hint for `$Id` here?

src/SAML2/XML/ds/KeyInfo.php
```diff
@@ -113,7 +86,7 @@ public function getInfo(): array
      * @param array $info
      * @return void
      */
-    public function setInfo(array $info): void
+    private function setInfo(array $info): void
     {
         $this->info = $info;
```
jaimeperez: We should probably have a check here to make sure the values in the array are only of the allowed classes (`\SAML2\XML\Chunk`, `\SAML2\XML\ds\KeyName` `\SAML2\XML\ds\X509Data`).

src/SAML2/XML/ds/KeyName.php
```diff
@@ -58,22 +51,33 @@ public function getName(): string
...
      * @return \DOMElement
      */
-    public function toXML(DOMElement $parent): DOMElement
+    public function toXML(DOMElement $parent = null): DOMElement
     {
-        Assert::notEmpty($this->name);
-
-        return Utils::addString($parent, XMLSecurityDSig::XMLDSIGNS, 'ds:KeyName', $this->name);
+//        $e = $this->instantiateParentElement($parent);
+        return Utils::addString($parent, self::NS, 'ds:KeyName', $this->name);
```
jaimeperez: Either we use `instantiateParentElement($parent)` and add the contents manually, or if we still want to use `Utils::addString()` we should change `'ds:KeyName'` to `static::NS_PREFIX.':'.static::class`.

src/SAML2/XML/ds/X509Certificate.php
jaimeperez: Maybe time for `SAML2\Utils::appendString()`? Then both `SAML2\Utils::addString()` and `SAML2\Utils::addStrings()` could be reimplemented to use that one, and here we could use it too.

src/SAML2/XML/ds/KeyInfo.php
tvdijen:
```
<element name="KeyInfo" type="ds:KeyInfoType"/>
- <complexType name="KeyInfoType" mixed="true">
- <choice maxOccurs="unbounded">
<element ref="ds:KeyName"/>
<element ref="ds:KeyValue"/>
<element ref="ds:RetrievalMethod"/>
<element ref="ds:X509Data"/>
<element ref="ds:PGPData"/>
<element ref="ds:SPKIData"/>
<element ref="ds:MgmtData"/>
<any processContents="lax" namespace="##other"/>
<!-- (1,1) elements from (0,unbounded) namespaces -->
</choice>
<attribute name="Id" type="ID" use="optional"/>
</complexType>
```

src/SAML2/XML/ds/KeyInfo.php
tvdijen: Agreed

src/SAML2/XML/ds/KeyName.php
tvdijen: I'll fiddle with it some more to see if I can get it to work

src/SAML2/XML/ds/KeyInfo.php
jaimeperez: The difference between `xsd:ID` and `xsd:string` is only semantical, if I'm not mistaken. In practice, an `xsd:ID` would always translate into a string for us.

src/SAML2/XML/ds/KeyName.php
tvdijen: This was fairly easy actually... Got it!

tvdijen:
Uhm, we don't return `null` from toXML in this PR?

jaimeperez:
My bad, I must have missed two open tabs. Does the excuse for lack of caffeine still work?

tvdijen:
That's always a valid excuse, but don't push it or you'll end up doing support again 🤣

PR #204, opened 2020-01-14
Migrate XML/mdui to new interface

tvdijen:


tvdijen:
I have again polluted this PR a bit with some things I missed earlier..

src/SAML2/XML/mdui/UIInfo.php
```diff
@@ -320,29 +362,45 @@ public function toXML(DOMElement $parent): ?DOMElement
...
+                foreach ($this->children as $child) {
+                    $child->toXML($e);
+                }
+            }
+
+            return $e;
         }
 
-        return $e;
+        return null;
```
jaimeperez: 😛

tvdijen: Taken care of, but still relies on #205

PR #205, opened 2020-01-14
Missed a few spots

tvdijen:
Some assertions and other stuff that we missed over the last few PRs

### DIFF
--- a/src/SAML2/XML/AbstractXMLElement.php
+++ b/src/SAML2/XML/AbstractXMLElement.php
@@ -32,7 +32,7 @@ abstract class AbstractXMLElement implements Serializable
     public function __toString(): string
     {
         $xml = $this->toXML();
-        return is_null($xml) ? '' : $xml->ownerDocument->saveXML($xml);
+        return $xml->ownerDocument->saveXML($xml);
     }
 
 
@@ -44,7 +44,7 @@ abstract class AbstractXMLElement implements Serializable
     public function serialize(): string
     {
         $xml = $this->toXML();
-        return is_null($xml) ? '' : $xml->ownerDocument->saveXML();
+        return $xml->ownerDocument->saveXML();
     }
 
 
@@ -104,7 +104,7 @@ abstract class AbstractXMLElement implements Serializable
      * Create XML from this class
      *
      * @param \DOMElement|null $parent
-     * @return \DOMElement|null
+     * @return \DOMElement
      */
-    abstract public function toXML(DOMElement $parent = null): ?DOMElement;
+    abstract public function toXML(DOMElement $parent = null): DOMElement;
 }
--- a/src/SAML2/XML/AbstractXMLElement.php
+++ b/src/SAML2/XML/AbstractXMLElement.php
@@ -91,6 +91,17 @@ abstract class AbstractXMLElement implements Serializable
 
 
     /**
+     * Test if an object, at the state it's in, would produce an empty XML-element
+     *
+     * @return bool
+     */
+    public function isEmptyElement(): bool
+    {
+        return false;
+    }
+
+
+    /**
      * Create a class from XML
      *
      * @param \DOMElement $xml
--- a/src/SAML2/XML/AbstractXMLElement.php
+++ b/src/SAML2/XML/AbstractXMLElement.php
@@ -32,7 +32,7 @@ abstract class AbstractXMLElement implements Serializable
     public function __toString(): string
     {
         $xml = $this->toXML();
-        return $xml->ownerDocument->saveXML($xml);
+        return is_null($xml) ? '' : $xml->ownerDocument->saveXML($xml);
     }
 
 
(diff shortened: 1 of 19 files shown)


tvdijen:
As of now I've been making the setters private, but the properties are protected..
Should we make the setters protected as well, and encourage people to use the setters if they're extending our classes?

jaimeperez:
It depends 😛

If the class is final, having protected properties doesn't serve any purpose because people won't be able to extend it. If the class is not final, then it might make sense, yes.

src/SAML2/XML/AbstractXMLElement.php
```diff
@@ -43,7 +43,8 @@ public function __toString(): string
      */
     public function serialize(): string
     {
-        return $this->toXML()->ownerDocument->saveXML();
+        $xml = $this->toXML();
+        return is_null($xml) ? '' : $xml->ownerDocument->saveXML();
```
jaimeperez: This won't work, and it might be a reason to actually return an empty element rather than `null`. The problem is that `unserialize()` relies on proper XML to reconstruct the object (via each object's `fromXML()` implementation). Take `AuthnContext` in this PR as an example. `toXML()` might return `null`. In that case, serializing leads to a string that contains no XML as its contents. When trying to unserialize later, `unserialize()` will call `fromXML()`, which requires a `DOMElement` passed as a parameter. This will fail with an `InvalidArgumentException` because the argument passed to `unserialize()` (`$serialized`, later in this file) will be an empty string, [which is explicitly disallowed](https://github.com/simplesamlphp/saml2/blob/master/src/SAML2/DOMDocumentFactory.php#L30-L32).

tvdijen: I see.. What about using '<root />' instead of the empty string.. Would that help?

jaimeperez: I don't think that would make any sense, to be honest. In any case, it would also fail since when processed into a `DOMElement`, the corresponding `fromXML()` method will check for the name of the tag, which will not match.

tvdijen: OK, changing this back to non-nullable again

jaimeperez: Sorry!

tvdijen: No worries! It's a puzzle we have to solve together..

I wonder if we can think of a clever way to determine whether the result of `$x->toXML($parent)` would be an empty tag.. I'm thinking along the lines of an isEmptyTag-method on our abstract class that will return false by default but can be overriden by the descending classes that would currently return `null`...

That would A) allow us to remove empty tags from message and B) still allow developers to work with empty tags as they desire.. What do you reckon?

jaimeperez: I was also thinking about that and I believe I have a solution. The thing is, it will only be some classes creating an empty tag, and in most cases we know if that'll be the case or not because we've created the object ourselves and know that we passed actual parameters to the constructor, and therefore the tag won't be empty.

For those cases were we can't know (e.g. the object was passed to us), there is actually a way to check if `toXML()` returned an empty tag:

```php
$e = $object->toXML();
if ($e->attributes->length === 0 && !$e->hasChildNodes()) {
    // empty tag
}
```

If we see this code repeating a lot, we could indeed create a simple static `isEmptyTag(DOMElement $e)` method in `AbstractXMLElement` or even `SAML2\Utils` so that we can just call that. What do you think?

tvdijen: I think we should go with a simple method and it doesn't have to be static.
``
if (!$object->isEmptyTag()) {
    $object->toXML($parent);
}
``
The isEmptyTag-method can be as easy as:
`return empty($this->optional1) && empty($this->optional2);` (or `return false;` as the default case)

jaimeperez: Oh, so you are thinking about checking this *before* actually calling `toXML()`. That's another option, although I don't really see much benefit (except for the save of a few CPU cycles creating the empty `DOMElement`). On the contrary, the downside I see with that approach is that those classes that can produce an empty element need to implement an extra method, while the post-`toXML()` approach allows you to do the same without extra code.

tvdijen: Agreed, we'll go with your solution

src/SAML2/XML/AbstractXMLElement.php
```diff
@@ -43,7 +43,8 @@ public function __toString(): string
      */
     public function serialize(): string
     {
-        return $this->toXML()->ownerDocument->saveXML();
+        $xml = $this->toXML();
+        return $xml->ownerDocument->saveXML();
```
jaimeperez: Psalm complaining? If that's the case, it worries me a bit, because getting as strict as we are getting with typing, psalm should be certain that the original statement wasn't going to fail...

tvdijen:
What about an optional second parameter to toXML to flag whether an empty element should be appended to the parent?

src/SAML2/XML/AbstractXMLElement.php
tvdijen: No, it doesn't complain.. I thought I'd mimic buildup of the __toString()-method

jaimeperez:
Great work Tim, and sorry again for the back and forth wrt testing for empty elements! 😓

tvdijen:
de nada!

jaimeperez:
¡Gracias! 😛

PR #207, opened 2020-01-21
WIP: Rewrite and migrate attribute-classes

tvdijen:


src/SAML2/XML/md/RequestedAttribute.php
```diff
@@ -8,36 +8,40 @@
...
-            return;
-        }
 
-        $this->isRequired = Utils::parseBoolean($xml, 'isRequired', null);
+    /**
+     * Initialize an RequestedAttribute.
+     *
+     * @param \SAML2\XML\saml\Attribute $attribute
+     */
+    public function __construct(Attribute $attribute, bool $isRequired = null)
```
tvdijen: I'm still in doubt whether I should pass an Attribute like I do now, or if I should pass the individual properties and create an Attribute-object in the constructor... I think the latter is the better option, because it would resemble the API of the Attribute-class

jaimeperez: I just had a look at the standard, and it indeed extends from `saml:Attribute`. Doing the same would make the implementation easier, because we would just need to add an extra boolean property (`isRequired`, defaulting to `null`), getter & setter, and set the namespace and prefix accordingly.

src/SAML2/XML/md/RequestedAttribute.php
```diff
@@ -58,26 +62,87 @@ public function getIsRequired(): ?bool
...
+
+        $friendlyName = $attribute->getFriendlyName();
+        if ($friendlyName !== null) {
+            $e->setAttribute('FriendlyName', $friendlyName);
+        }
+
+        $attributeValues = $attribute->getAttributeValues();
+        if (!empty($attributeValues)) {
+            foreach ($attributeValues as $av) {
+                $e->appendChild($e->ownerDocument->importNode($av->toXML(), true));
```
jaimeperez: Shouldn't this be $av->toXML($e)?

src/SAML2/XML/saml/Attribute.php
```diff
@@ -129,65 +133,54 @@ public function getFriendlyName(): ?string
...
-    {
-        $this->setAttributeValues(
-            empty($this->attributeValues) ? [$attributeValue] : array_merge($this->attributeValues, [$attributeValue])
-        );
+        if ($attributeValues === null) {
+            return;
+        }
+        Assert::allIsInstanceOf($attributeValues, AttributeValue::class, 'Invalid AttributeValue.');
+        $this->AttributeValues = $attributeValues;
     }
```
tvdijen: Why are you turning an adder-method into a setter-method?

src/SAML2/XML/saml/Attribute.php
```diff
@@ -129,65 +133,54 @@ public function getFriendlyName(): ?string
...
-
     /**
      * Convert XML into a Attribute
      *
-     * @param \DOMElement $xml The XML element we should load
-     * @return \SAML2\XML\saml\Attribute
+     * @param DOMElement $xml The XML element we should load
+     *
+     * @return Attribute
+     * @throws Exception
```
tvdijen: Is this your IDE interfering? We use full namespaces in phpdoc throughout this project..
It's especially helpful in really large files, so you don't have to scroll all the way to the top to find out in what namespace `Attribute` exists..

src/SAML2/XML/saml/Attribute.php
```diff
@@ -223,8 +216,8 @@ public function toXML(DOMElement $parent = null): DOMElement
             $e->setAttribute('FriendlyName', $this->FriendlyName);
         }
 
-        if (!empty($this->attributeValues)) {
-            foreach ($this->attributeValues as $av) {
+        if (!empty($this->AttributeValuess)) {
```
tvdijen: That'sssssss a typo!

jaimeperez: Yep, fixed it in the other branch.

src/SAML2/XML/saml/Attribute.php
jaimeperez: Yeah, it marks those as "_unnecessarily qualified_". I can revert those changes, having full namespaced names makes sense in the phpdoc 👍

Thanks for the heads up!

src/SAML2/XML/saml/Attribute.php
jaimeperez: I'm not, I just removed the adder, since I'm building the entire set manually and calling the setter. Having immutable objects, there's no point in having the adder, so I just removed it as in other places of the code.

tvdijen: My bad..  BTW, I didn't notice you had merged this one into the other PR.. Is it still useful to keep this one open?

jaimeperez: I'll leave it up to you, as I don't know if you were planning to do any more work on this, or it works fine as it is now. Considering that there are merge conflicts, it might make sense to just drop it if there's nothing else we want to add to this one...

PR #208, opened 2020-01-27
WIP Rewite the API provided by all classes under XML/md/

jaimeperez:
This is quite a big piece of work. Some details:

- [x] Rewrite all classes under `XML/md/` to be immutable.
- [x] Redesign the signature-related classes to be interfaces and traits that can be used in whatever element, without messing with the inheritance relationships.
- [x] Add new classes to model missing `EndpointType` extensions.
- [x] Add new classes to model missing `LocalizedNameType` extensions.

Some notes on the approach followed:

* Some classes are not abstract, but we want them to be extensible as well. Since we need `fromXML()` methods for both, the classes extending from the base cannot reuse `fromXML()` since that one checks the actual element name and namespace URI. An alternative is to create static methods for each property, that return any valid value for it, from a given XML. These methods include all checks needed (e.g. a `anyURI` type cannot contain an empty string).

* I've used traits to achieve composition of functionality common between different classes. When appropriate, I've added meaningful abstract classes that aggregate functionality coming from those traits (e.g. an `AbstractMetadataDocument` class that adds properties –and methods– related to any SAML metadata document, like `cacheDuration` or `validUntil`).


src/SAML2/SignedElementHelper.php
```diff
@@ -15,54 +16,41 @@
...
     /**
-     * Available methods for validating this message.
-     *
-     * @var array
-     */
-    private $validators = [];
-
-    /**
-     * How long this element is valid, as a unix timestamp.
+     * SignedElement constructor.
```
tvdijen: This is not a constructor?

src/SAML2/SignedElementHelper.php
```diff
@@ -15,54 +16,41 @@
...
      *
-     * @var int|null
+     * @param XMLSecurityDSig|null $signature
```
tvdijen: @var was correct here..   And again, I think your IDE is 'optimizing' the namespaces away, but it's not what we do throughout the rest of the project

src/SAML2/XML/saml/Attribute.php
```diff
@@ -119,12 +106,15 @@ public function getNameFormat(): ?string
...
-     * @param string|null $nameFormat
-     * @return void
+     * @param string|null $NameFormat
      */
-    public function setNameFormat(string $nameFormat = null): void
+    protected function setNameFormat(?string $NameFormat): void
     {
-        $this->NameFormat = $nameFormat;
+        if ($NameFormat === null) {
+            return;
```
tvdijen: It may be useful to actually set it to `null`, so I would suggest to negate the if-statement and move the assertion in there..

src/SAML2/XML/saml/Attribute.php
```diff
@@ -143,67 +133,79 @@ public function getFriendlyName(): ?string
...
-     * @param array $attributeValue
-     * @return void
+     * @param array $attributeValues|null
      */
-    public function setAttributeValue(array $attributeValue): void
+    protected function setAttributeValues(?array $attributeValues): void
     {
-        $this->AttributeValue = $attributeValue;
+        if ($attributeValues === null) {
+            return;
```
tvdijen: Same

src/SAML2/SignedElementHelper.php
jaimeperez: Good catch! I don't know what happened here...

tvdijen: Lack of coffee is what happened :P

src/SAML2/XML/saml/Attribute.php
jaimeperez: It defaults to `null`, so in case `null` is passed, `$this->NameFormat` was already `null`. Remember that the setters are intended only to set properties with all sanity checks needed during construction.

tvdijen: But the class is not final, so it could potentially be called more than once from an extending class..
I know it's not very likely to happen, but I'd like to keep things consistent and I know we have a couple of cases where we do as I described..

tvdijen: One of them:
https://github.com/simplesamlphp/saml2/pull/197/files#diff-5068cfce0ab6b548e1e6f9f7818e97afR158

src/SAML2/SignedElementHelper.php
jaimeperez: Amen, brother!

src/SAML2/XML/saml/Attribute.php
jaimeperez: Uhm, that's a good point! That actually raises a bigger concern. We're removing checks in `toXML()` since the object is guaranteed to be properly created at all times. But if a class extends it, the properties are protected and can be accessed by the extending class, therefore breaking all guarantees for immutability.

I'm thinking we need `SAML2\XML\saml\Attribute` to be declared `final`. We can do that and keep `md\RequestedAttribute` the way it is by either moving all the functionality in `Attribute` into a trait, or by creating an abstract `SAML2\XML\saml\AbstractAttributeType` class, and make both `Attribute` and `RequestedAttribute` extend it.

What do you think?

src/SAML2/SignedElementHelper.php
```diff
@@ -9,60 +9,44 @@
...
 /**
- * Helper class for processing signed elements.
- *
- * Can either be inherited from, or can be used by proxy.
+ * Helper trait for processing signed elements.
  *
  * @package SimpleSAMLphp
  */
-class SignedElementHelper extends SignedElement
+trait SignedElementHelper
```
tvdijen: If we want to follow PSR naming conventions, we should suffix this class with 'Trait' and rename the file accordingly

src/SAML2/XML/AbstractXMLElement.php
```diff
@@ -43,7 +44,7 @@ public function __toString(): string
      */
     public function serialize(): string
     {
-        return $this->toXML()->ownerDocument->saveXML();
+        return $this->__toString();
```
tvdijen: I have reverted this line, because it was breaking tests:
https://travis-ci.org/simplesamlphp/saml2/jobs/643135861#L329

src/SAML2/SignedElementHelper.php
jaimeperez: True! I was thinking about that, but was uncertain whether the pattern was followed for traits as well...

src/SAML2/XML/AbstractXMLElement.php
jaimeperez: Uhm, it's true that the string representation is formatted, while the serialized one should not. Still, in this case I think we should avoid the document type declaration at the top, and this way we won't, since we are not specifying the root element to print.

src/SAML2/SignedElementHelper.php
tvdijen: It is, and I have fixed this in the simplification-PR, so basically this is _Resolved_ now..
References to traits need to be fixed in this PR though..

src/SAML2/XML/AbstractXMLElement.php
tvdijen: OK so I've basically made a copy of toString, but without the formatting line... Is that what you were saying?

tvdijen:
I was able to merge quite a good portion of this PR to master without breaking anything.. That way we at least have some of the individual commits + commit messages instead of one huge squashed commit.

src/SAML2/XML/md/AbstractIndexedEndpointType.php
```diff
@@ -87,7 +87,7 @@ public static function fromXML(DOMElement $xml): object
      */
     public function toXML(DOMElement $parent = null): DOMElement
     {
-        Assert::notEmpty($this->index);
+        Assert::numeric($this->index);
```
tvdijen: May I suggest s/numeric/natural ?

jaimeperez: You may, but we can't use naturals there 😛

Naturals don't include 0, but that's a valid index number, AFAIU.

tvdijen: 0 is included in webmozart's natural assertion 😉

jaimeperez: Uhm, didn't know that webmozart behaves like that, but I just found (to my surprise) that some people (even standards) consider naturals to start at 0... 😮

In any case, agreed, the test should be in the setter. I'll fix it.

tvdijen: Yeah, apparently there is no general consensus on whether 0 is a natural number or not, although in most cases it is excluded.. Webmozart specifically defines it as 'non-negative'..

tvdijen:
Thanks, I had to run into a meeting :P

jaimeperez:
No prob! 😄

src/SAML2/SignedElementTrait.php
```diff
@@ -93,28 +117,23 @@ public function addValidator(callable $function, $data): void
...
 
-                return true;
-            } catch (Exception $e) {
-                $exceptions[] = $e;
-            }
         }
-
-        // no validators were able to validate the message.
-        throw $exceptions[0];
+        throw new \Exception("Unable to validate Signature");
```
tvdijen: I was always taught to avoid the built-in \Exception ...
Should we use one of our own exceptions here, like RuntimeException?

jaimeperez: Maybe even a specific exception for signature validation errors...

tvdijen: I'm glad we all agree :+1:

PR #209, opened 2020-01-28
Simplify #208

tvdijen:
This PR contains some of the commits from #208 that are stand-alone changes, so we can undress that one and make it easier to review.

Some notes:
- I kept the original SignedElementHelper for now; We can remove it in #208 as soon as all classes are migrated to use the new SignedElementTrait.
- The Traits have been renamed to follow PSR naming conventions
- One line has been reverted because it was causing issues with the tests;  see comments on #208


tvdijen:
I think Scrutinizer has some valid remarks here:
https://scrutinizer-ci.com/g/simplesamlphp/saml2/inspections/47e8b565-1df0-4e59-85de-900722411f91

src/SAML2/XML/AbstractXMLElement.php
```diff
@@ -101,6 +105,138 @@ public function isEmptyElement(): bool
...
+
+
+    /**
+     * Extract localized names from a the children of a given element.
+     *
+     * @param DOMElement $parent The element we want to search.
+     *
+     * @return array An array of objects of class $class.
+     */
+    public static function extractFromChildren(\DOMElement $parent): array
```
jaimeperez: I'm still in doubt about this method's name. What do you think about something like `getChildrenOfType()`?

src/SAML2/XML/ExtendableElementTrait.php
```diff
@@ -0,0 +1,140 @@
...
+use SAML2\XML\mdui\DiscoHints;
+use SAML2\XML\mdui\UIInfo;
+use SAML2\XML\shibmd\Scope;
+
+/**
+ * Trait for metadata elements that can be extended.
+ *
+ * @package simplesamlphp/saml2
+ */
+trait ExtendableElementTrait
```
jaimeperez: I changed this one yesterday. Basically, the entire way extensions were implemented was overcomplicated and nonsensical. I tried to simplify everything by having `md:Extensions` do all the heavy lifting in the two standard `fromXML()` and `toXML()` methods (it now extends `AbstractXMLElement`), and then this trait just adds the property with the `SAML2\XML\md\Extensions` object and the common getter and setter methods.

src/SAML2/XML/md/AbstractMetadataDocument.php
```diff
@@ -0,0 +1,212 @@
...
+
+
+    /**
+     * Process an XML element and get its ID property, if any.
+     *
+     * @param \DOMElement $xml An element that may contain an ID.
+     *
+     * @return string|null
+     */
+    public static function getIDFromXML(DOMElement $xml): ?string
```
jaimeperez: We should get rid of this one and use `self::getAttribute()` instead.

src/SAML2/XML/md/AbstractMetadataDocument.php
```diff
@@ -0,0 +1,212 @@
...
+    /**
+     * Process an XML element and get its validUntil property, if any.
+     *
+     * @param \DOMElement $xml An element that may contain validUntil.
+     *
+     * @return int|null
+     *
+     * @throws \Exception If the processed validUntil from $xml is not a valid timestamp.
+     */
+    public static function getValidUntilFromXML(DOMElement $xml): ?int
```
jaimeperez: This one I already got rid of.

src/SAML2/XML/md/AbstractMetadataDocument.php
```diff
@@ -0,0 +1,212 @@
...
+
+
+    /**
+     * Process an XML element and get its cacheDuration property, if any.
+     *
+     * @param \DOMElement $xml An element that may contain cacheDuration.
+     *
+     * @return string|null
+     */
+    public static function getCacheDurationFromXML(DOMElement $xml): ?string
```
jaimeperez: And this one should disappear as well.

tests/SAML2/XML/saml/AttributeValueTest.php
```diff
@@ -95,9 +95,8 @@ public function testCreateAttributeFromDOMElement(): void
     public function testSerialize(): void
     {
         $av1 = new AttributeValue("Aap:noot:mies");
-        $ser = $av1->serialize();
         $av2 = new AttributeValue("Wim");
-        $av2->unserialize($ser);
+        $av2 = unserialize(serialize($av1));
 
         $this->assertEquals("Aap:noot:mies", $av2->getString());
```
jaimeperez: I tried to write these tests a bit simpler. If we already have a DOMDocument in the test class, we can just load it with `fromXML()`, serialize, unserialize, and then compare the string value with the result of calling `saveXML()` on the document.

tests/SAML2/XML/saml/AuthnContextTest.php
```diff
@@ -69,10 +74,7 @@ public function testMarshallingWithoutClassRef(): void
         );
 
         $document = DOMDocumentFactory::fromString('<root />');
-        /**
-         * @var \DOMElement $document->firstChild
-         * @var \DOMElement $authnContextElement
-         */
+        /** @var \DOMElement $document->firstChild */
         $authnContextElement = $authnContext->toXML($document->firstChild);
```
jaimeperez: This should use `$document->documentElement`, since `firstChild` is not guaranteed to be a `DOMElement`.

src/SAML2/XML/AbstractXMLElement.php
tvdijen: I would keep it as is, because 'Type' may raise confusion with SAML types.. I do see the phpdoc for this method needs a polish!

jaimeperez: What about `getChildrenOfClass()`?

tvdijen: Renamed & touched up the method

src/SAML2/XML/ExtendableElementTrait.php
tvdijen: I have indeed missed that due to the filename/classname change.. Fixed

src/SAML2/XML/md/AbstractMetadataDocument.php
tvdijen: Removed

src/SAML2/XML/md/AbstractMetadataDocument.php
tvdijen: Removed

src/SAML2/XML/md/AbstractMetadataDocument.php
tvdijen: Removed

tests/SAML2/XML/saml/AttributeValueTest.php
tvdijen: I think it has some ties to the rewrite-attributes PR, so we should probably fix it there..

tests/SAML2/XML/saml/AuthnContextTest.php
tvdijen: Done

src/SAML2/XML/ExtendableElementTrait.php
tvdijen: Something's off in this file with the Extensions-property.. phpdoc says it's an array of extensions, but we're not treating it that way.. What did you intent to do?

PR #210, opened 2020-01-29
Add mailto:-prefix in EmailAddress property, if missing.

thijskh:
This is mandatory in the spec. It seems the role of the library to
add it, so users can continue to supply and receive bare email
addresses. The code is defensive in that it will also accept
incoming EmailAddress values that already sport the mailto:-prefix.

Also improve code coverage of ContactPerson

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# Editor configuration normalization
+# @see http://editorconfig.org/
+
+# This is the top-most .editorconfig file; do not search in parent directories.
+root = true
+
+# All files.
+[*]
+end_of_line = LF
+indent_style = space
+indent_size = 4
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
--- a/src/SAML2/XML/md/ContactPerson.php
+++ b/src/SAML2/XML/md/ContactPerson.php
@@ -266,7 +266,7 @@ class ContactPerson
      */
     private function removeEmailMailtoPrefix(string $emailAddress): string
     {
-        return substr(strtolower($emailAddress),0,7) === 'mailto:' ? substr($emailAddress,7) : $emailAddress;
+        return preg_replace('/^mailto:/i', '', $emailAddress);
     }
 
     /**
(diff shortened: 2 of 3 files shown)


thijskh:
@msalle

tvdijen:
Good catch! Should probably be backported to 3x. and 4.x as well..

I do wonder if we can reduce this to 1) always remove mailto: when setting the value and 2) always insert mailto: when generating the XML?

msalle:
thanks! As far as I can see, you'll also need to update [lib/SimpleSAML/Utils/EMail.php](https://github.com/simplesamlphp/simplesamlphp/blob/master/lib/SimpleSAML/Utils/EMail.php), since that uses the bare value from the config, via [getDefaultMailAddress()](https://github.com/simplesamlphp/simplesamlphp/blob/master/lib/SimpleSAML/Utils/EMail.php#L65).

thijskh:
I don't think we need to update that class, since when sending email it is exactly right to be using the bare email address? The problem is just the representation in metadata which requires the extra `mailto:`.

msalle:
True, it's not necessary if people use the bare email address in the technical contact information, but since that broke the metadata before this PR, it will typically now include a mailto: which breaks the email functionality. Since you have a nice function to get the bare address, isn't it nicer to use it there too?

thijskh:
That is indeed the most backward compatible, yes.

thijskh:
> Should probably be backported to 3x. and 4.x as well..

Since it's more or less (!) bc that should be fine indeed.

> I do wonder if we can reduce this to 1) always remove mailto: when setting the value and 2) always insert mailto: when generating the XML?

Is also a good solution, depends a bit on what you'd expect the internal representation to be I guess. So I'm not sure either is "better"?

jaimeperez:
I agree with Tim. An email should always be an email, and the `mailto:` prefix is a detail of its specific SAML representation. Therefore, it makes sense to store bare email addresses, strip the prefix when setting it (if present), and add it on `toXML()`.

thijskh:
Alright, changed it that way, ready to merge?

tvdijen:
I hope you don't mind, but I've polished it up a bit..
I'm thinking we should filter_var the values as well, making sure they are valid email addresses..
Other than that 👍

thijskh:
Sure, my approach is that I tend to avoid preg when there’s no actual regular expressions, just static strings involved. But that’s likely just a personal matter of preference. So this is fine indeed, it is more compact.

Checking incoming email for validity in the setters is a good idea but maybe separately because it has a slight chance of breaking existing stuff, so perhaps do that only for master

PR #213, opened 2020-02-02
Rewrite api saml

tvdijen:
~~@jaimeperez I need some help on this one.. The NameID/Issuer part is rather confusing~~


tvdijen:
@jaimeperez I need some help on this one.. The NameID/Issuer part is rather confusing

tvdijen:
One test is failing right now and I think it's a matter of how you read the specs..

> 2.2.5 Element <Issuer>The <Issuer> element, with complex type NameIDType, provides information about the issuer of aSAML assertion or protocol message. The element requires the use of a string to carry the issuer's name, but permits various pieces of descriptive data (see Section 2.2.2).
> Overriding the usual rule for this element's type, if no Format value is provided with this element, then the value urn:oasis:names:tc:SAML:2.0:nameid-format:entity **is in effect** (see Section 8.3.6).

The way I read 'in effect' is that we have to assume the entity format is being used when processing incoming messages.. Not that we have to include it in outgoing messages, because that would mean we can never omit an attribute that's supposed to be [Optional] by spec..

@thijskh @jaimeperez what do you think?

tvdijen:
The way SPID-support was implemented is probably not the best.. We may want to consider creating `src/XML/spid/Issuer.php` to take care of this unusual case..
It also seems we do more than they actually need.. Only NameQualifier is required by SPID.. The others are not.
https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=3&ved=2ahUKEwjKj_f6uLPnAhWJJlAKHaGLAqgQFjACegQIBRAB&url=https%3A%2F%2Fec.europa.eu%2Fcefdigital%2Fwiki%2Fdownload%2Fattachments%2F62885733%2FT.1%2520-%2520Regulation%2520setting%2520out%2520the%2520technical%2520rules.pdf%3Fversion%3D1%26modificationDate%3D1531759228950%26api%3Dv2&usg=AOvVaw0cRt6zACjcOx7QwS0oLbWu

src/SAML2/XML/saml/NameIDType.php
```diff
@@ -55,27 +55,31 @@ abstract class NameIDType extends BaseIDType
...
-        parent::__construct($xml);
-
-        if ($xml === null) {
-            return;
-        }
-
-        if ($xml->hasAttribute('Format')) {
-            $this->Format = $xml->getAttribute('Format');
+    public function __construct(
+        $element,
```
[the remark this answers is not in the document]
tvdijen: Beats me.. I think I may have migrated this wrong.. Will look into it..
As far as I can tell, the only thing SPID _really_ needs is to have SPNameQualifier exposed on this Issuer.. The advantage of an extension would be that we can also add assertions for the other SPID-specific characteristics

src/SAML2/XML/saml/NameID.php
```diff
@@ -13,9 +16,30 @@
...
+     * @return \SAML2\XML\saml\NameID
+     * @throws \Exception
      */
-    protected $nodeName = 'saml:NameID';
+    public static function fromXML(DOMElement $xml): object
+    {
+        Assert::same($xml->localName, 'NameID');
+        Assert::same($xml->namespaceURI, NameID::NS);
+
+        $Format = $xml->hasAttribute('Format') ? $xml->getAttribute('Format') : null;
```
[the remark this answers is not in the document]
tvdijen: Yes, good catch! There may be more cases like this throughout the XML-dir.. I'll scan for it...

src/SAML2/XML/saml/NameIDType.php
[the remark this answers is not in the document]
tvdijen: It was in fact a migration issue on my end..

src/SAML2/XML/saml/NameIDType.php
```diff
@@ -96,7 +95,7 @@ public function getFormat(): ?string
      * @param string|null $format
      * @return void
      */
-    public function setFormat(string $format = null): void
+    public function setFormat(?string $format): void
```
jaimeperez: Protected?

Also, is the format supposed to be an URI? Because in that case we should probably discard empty values at least.

src/SAML2/XML/saml/NameIDType.php
```diff
@@ -146,12 +144,33 @@ public function getSPProvidedID(): ?string
...
+    public static function fromXML(DOMElement $xml): object
+    {
+        Assert::NotEmpty($xml->textContent);
+
+        $Format = self::getAttribute($xml, 'Format', null);
+        $SPProvidedID = self::getAttribute($xml, 'SPProvidedID', null);
+        $NameQualifier = self::getAttribute($xml, 'NameQualifier', null);
+        $SPNameQualifier = self::getAttribute($xml, 'SPNameQualifier', null);
+
+        return new static($xml->textContent, $Format, $SPProvidedID, $NameQualifier, $SPNameQualifier);
```
jaimeperez: This is tricky (been there, done that). Any class extending `NameIDType` needs to respect this exact signature of the constructor. We should document this in the phpdoc.

src/SAML2/XML/saml/SubjectConfirmation.php
```diff
@@ -138,7 +138,7 @@ public static function fromXML(DOMElement $xml): object
         Assert::same($xml->namespaceURI, SubjectConfirmation::NS);
 
         if (!$xml->hasAttribute('Method')) {
-            throw new \Exception('SubjectConfirmation element without Method attribute.');
+            throw new \InvalidArgumentException('SubjectConfirmation element without Method attribute.');
```
jaimeperez: This one can also be expressed with webmozart, if we want (I've done that in some places):

```php
        Assert::true($xml->hasAttribute('Method'), 'SubjectConfirmation element without Method attribute.');
```

src/SAML2/XML/saml/SubjectConfirmationData.php
```diff
@@ -60,9 +60,9 @@ final class SubjectConfirmationData extends AbstractSamlElement
      * Array with various elements describing this key.
      * Unknown elements will be represented by \SAML2\XML\Chunk.
      *
-     * @var (\SAML2\XML\ds\KeyInfo|\SAML2\XML\Chunk)[]|null
+     * @var (\SAML2\XML\ds\KeyInfo|\SAML2\XML\Chunk)[]
      */
-    private $info = null;
+    private $info = [];
```
jaimeperez: This needs to be protected, or serialization won't work.

src/SAML2/XML/saml/SubjectConfirmationData.php
```diff
@@ -226,10 +226,10 @@ public function getInfo(): ?array
...
      * Set the value of the info-property
      *
-     * @param (\SAML2\XML\ds\KeyInfo|\SAML2\XML\Chunk)[]|null $info
+     * @param (\SAML2\XML\ds\KeyInfo|\SAML2\XML\Chunk)[] $info
      * @return void
      */
-    private function setInfo(?array $info): void
+    private function setInfo(array $info): void
     {
         $this->info = $info;
```
jaimeperez: We should have a check here:

```php
        Assert::allIsInstanceO(Chunk::class, 'All info elements must be an instance of SAML2\XML\Chunk.');
```

src/SAML2/XML/saml/SubjectConfirmationData.php
```diff
@@ -246,7 +246,7 @@ private function setInfo(?array $info): void
     public function addInfo(object $info): void
```
jaimeperez: In the other branch I've been getting rid of all adders. If we are going for immutability (even at the expense of hunge constructor signatures), then I think we shouldn't allow public methods that modify the state like this.

tests/SAML2/XML/spid/IssuerTest.php
```diff
@@ -0,0 +1,56 @@
...
+class IssuerTest extends TestCase
+{
+    /**
+     * @return void
+     */
+    public function testMarshalling(): void
+    {
+        $issuer = new Issuer('TheIssuerValue', Constants::NAMEID_ENTITY, 'TheSPProvidedID', 'TheNameQualifier', 'TheSPNameQualifier');
+        $issuerElement = $issuer->toXML();
+        $issuerElements = Utils::xpQuery($issuerElement, '/saml_assertion:Issuer');
```
jaimeperez: I find it easier to have an XML document built from a string (which can be done in `setUp()`, and then compare that one with the string generated by the object we just created. That way we don't need to be manipulating the DOM in order to test if it's correct. Actually, such comparison allows us to verify that what we generate is exactly what we expect, while with this approach we could be missing things because the single tests we are doing are fine.

An example from today: we had a bug in one class (I think `SPSSODescriptor`) that made the `md:Extensions` object be duplicated into the resulting DOM. Using an approach like this, I wouldn't have noticed the problem. However, comparing the exact resulting string uncovered it.

tests/SAML2/XML/spid/IssuerTest.php
```diff
@@ -0,0 +1,56 @@
...
+XML
+        );
+
+        $issuer = Issuer::fromXML($document->firstChild);
+        $this->assertEquals('TheNameQualifier', $issuer->getNameQualifier());
+        $this->assertEquals('TheSPNameQualifier', $issuer->getSPNameQualifier());
+        $this->assertEquals(Constants::NAMEID_ENTITY, $issuer->getFormat());
+        $this->assertEquals('TheSPProvidedID', $issuer->getSPProvidedID());
+        $this->assertEquals('TheIssuerValue', $issuer->getValue());
+    }
```
jaimeperez: We should also have tests for serialization. Sometimes we miss variables that were private, and then serialization works. The only way to find out is testing serialization / unserialization of every object.

src/SAML2/XML/saml/SubjectConfirmationData.php
tvdijen: I thought you wanted to keep them? Have you changed your mind?

src/SAML2/XML/saml/NameIDType.php
tvdijen: I think in this case it is fairly safe, but I don't mind duplicating four lines across Issuer/NameID as well... What would you prefer?

tvdijen: I decided to go safe and get rid of the static call at the cost of 4~5 lines code duplication..

src/SAML2/XML/saml/NameIDType.php
```diff
@@ -96,59 +97,59 @@ public function getFormat(): ?string
      * @param string|null $format
      * @return void
      */
-    public function setFormat(string $format = null): void
+    private function setFormat(?string $format): void
     {
+        Assert::nullOrNotWhitespaceOnly($format);
```
jaimeperez: I should be using this one in `md/` as well...

src/SAML2/XML/saml/SubjectConfirmationData.php
jaimeperez: Yeah. I panicked first when I saw the monster signatures of some constructors, but then thought it was stupid to aim for immutability if that was going away with adders. I think we discussed this at some point, and agreed to make everything strictly immutable at least in the first iteration, and revisit it later.

src/SAML2/XML/spid/Issuer.php
```diff
@@ -0,0 +1,69 @@
...
+use SAML2\XML\saml\NameIDType;
+use Webmozart\Assert\Assert;
+
+/**
+ * Class representing the saml:Issuer element compliant with SPID spefication.
+ *
+ * @author Tim van Dijen, <tvdijen@gmail.com>
+ * @package simplesamlphp/saml2
+ */
+final class Issuer extends NameIDType
```
jaimeperez: Does this element use the SAML metadata namespace, according to the SPID specification? I'm wondering if that's allowed, because they would essentially be overriding the original schema.

If that's the case, I wonder as well how can we process these elements if we receive an XML with a regular `md:Issuer` element that follows the SPID specification instead of the SAML standard....

tvdijen: They just follow the namespace for SAML :(

src/SAML2/XML/saml/SubjectConfirmationData.php
tvdijen: I think the aim was to make sure no illegal situations can occur.. Not per se immutability?

src/SAML2/XML/spid/Issuer.php
tvdijen: Maybe we should not enforce too much of the spec in the library and only focus on XML structural integrity and presence of mandatory fields..
Business-logic like this can then go into SSP, together with SAML2int rules..

jaimeperez: There's a third approach for cases like this, which is completely ignoring the issue during parsing. As in: "_there shouldn't be this field here, let's just ignore it then_". Now, that would work on the IdP side. We could essentially ignore the attribute to the point of not even setting the property internal to the `Issuer` object. That way a SimpleSAMLphp will at least not choke with an authentication request coming from an SPID service provider. If you really want to get the `NameQualifier` attribute (which you shouldn't be using), well, you can always fetch it manually.

Now, from the SP point of view, I think it's important that messages produced by our library don't break the standard. I can live with this being done in a separate class, because then it's the responsibility of the service to break it consciously. In any case, the SPID implementation of `Issuer` can be greatly simplified, because the format is fixed, and only the `NameQualifier` and the value are required.

There's a third case, which is an IdP sending a response with an assertion. Assertions are fine, but if we look at the example response, the `Issuer` there contains both `Format` and `NameQualifier`. However, the text describing the `Issuer` element of a response doesn't say anything about those:

> the element &lt;Issuer&gt;must be present, indicating the entityID of the issuing entity, i.e. the Identity Provider itself; the attribute format must be omitted or have the value “urn:oasis:names:tc:SAML:2.0:nameid-format:entity”;

I'm therefore inclined to ignore the example and assume this was a mistake by the ones writing the spec. That would leave us with the custom, very basic class for authentication requests, and a relaxed test for everything else, ignoring any extra attributes if the format is _entity_. Would that be reasonable?

tvdijen: Not sure I follow.. Would it be to much to ask to just do add your last suggestion to this PR? It sound like a minor change.

jaimeperez: My response was way more convoluted than the code I'd suggest 🤣

I'll do it 👍

jaimeperez: There, have a look!

tvdijen:
Isn't this basically what I said? Loosen the checks on the library and enforce them on SSP? I wonder if we shouldn't get rid of the spid-class completely..

tvdijen:
Just a final sanity check on the last commit please?

tvdijen:
Just a final sanity check on the last two commits please?

PR #215, opened 2020-02-05
Sanitize alg: tests

tvdijen:
Add serialization tests as well

### DIFF
--- a/tests/SAML2/XML/alg/DigestMethodTest.php
+++ b/tests/SAML2/XML/alg/DigestMethodTest.php
@@ -38,6 +38,9 @@ final class DigestMethodTest extends \PHPUnit\Framework\TestCase
     public function testMarshalling(): void
     {
         $digestMethod = new DigestMethod('http://exampleAlgorithm');
+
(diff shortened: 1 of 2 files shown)


tests/SAML2/XML/alg/DigestMethodTest.php
```diff
@@ -15,25 +15,30 @@
...
-        $this->assertCount(1, $digestMethodElements);
-        $digestMethodElement = $digestMethodElements[0];
-        $this->assertEquals('http://exampleAlgorithm', $digestMethodElement->getAttribute('Algorithm'));
+    /**
+     * @return void
+     */
+    public function testMarshalling(): void
+    {
+        $digestMethod = new DigestMethod('http://exampleAlgorithm');
+        $this->assertEquals($this->document->saveXML($this->document->documentElement), strval($digestMethod));
```
jaimeperez: I typically add tests for all the getters in the object, so that we not only test that `toXML()` works fine, but also that we can extract the information properly from the object.

PR #220, opened 2020-02-25
Rename classes and move them to the right namespace

tvdijen:
No functional changes


src/SAML2/HTTPArtifact.php
```diff
@@ -169,7 +172,7 @@ public function receive(): Message
             throw new \Exception('Empty ArtifactResponse received, maybe a replay?');
         }
 
-        $samlResponse = Message::fromXML($xml);
+        $samlResponse = AbstractMessage::fromXML($xml);
```
jaimeperez: This is the method we agreed to move to an standalone factory.

src/SAML2/XML/samlp/AbstractMessage.php
```diff
@@ -431,8 +420,10 @@ public function setRelayState(string $relayState = null): void
      *
      * @return \DOMElement The root element of the DOM tree
      */
-    public function toXML(): DOMElement
+    public function toXML(?DOMElement $parent = null): DOMElement
     {
+        Assert::null($parent);
```
jaimeperez: I think there can be situations where a message is not the actual root of the XML document (SOAP? PAOS?), so we should actually use `$parent` here.

src/SAML2/XML/samlp/AbstractStatusResponse.php
```diff
@@ -132,16 +132,18 @@ public function setStatus(Status $status): void
      *
      * @return \DOMElement This status response.
      */
-    public function toXML(): DOMElement
+    public function toXML(?DOMElement $parent = null): DOMElement
     {
-        $root = parent::toXML();
+        Assert::null($parent);
```
jaimeperez: Same here.

src/SAML2/XML/samlp/AbstractSubjectQuery.php
```diff
@@ -109,17 +112,18 @@ public function setNameId(NameID $nameId): void
      *
      * @throws \InvalidArgumentException if assertions are false
      */
-    public function toXML(): DOMElement
+    public function toXML(?DOMElement $parent = null): DOMElement
     {
+        Assert::null($parent);
```
jaimeperez: And here.

tvdijen:
As discussed the comments will be addressed in the next PR where we migrate the classes to the new API

PR #223, opened 2020-02-28
Fix Attribute + SubjectConfirmationData

tvdijen:
According to the specs these elements can contain
`<anyAttribute namespace="##other" processContents="lax"/>`


tvdijen:
Thanks! I see there are two Scrutinizer-issues that need to be addressed though

tvdijen:
They turned out to be false positives, so merging..

PR #224, opened 2020-02-29
Add class registry for extensions

tvdijen:


src/SAML2/Compat/ContainerSingleton.php
```diff
@@ -45,4 +46,25 @@ public static function initSspContainer(): Container
...
+
+
+    /**
+     * @param string $className
+     * @param array $qualifiedName
+     * @return void
+     */
+    public static function registerClass(string $className, array $qualifiedName): void
+    {
+        $this->registry[$qualifiedName] = $className;
```
jaimeperez: Is it possible to index an array with... array elements?

I think it's easier (and clearer) if we specify separately the namespace and the name of the element. Note also that the name should **not** be the qualified name, as the prefix is irrelevant in this case.

src/SAML2/Compat/ContainerSingleton.php
```diff
@@ -45,4 +46,25 @@ public static function initSspContainer(): Container
...
+    }
+
+
+    /**
+     * @param string $className
+     * @return array|false
+     */
+    public static function getRegisteredClass(string $className)
+    {
+        return array_search($className, $this->registry);
```
jaimeperez:
```suggestion
    public static function getRegisteredClass(string $className)
    {
        return array_search($className, $this->registry);
```

What we need to search for is the namespace and element name. We need the registry to be able to run `fromXML()` on arbitrary classes we don't know about on beforehand. The only piece of information that we have when we need to call those classes is the namespace and the name of an elementt we want to process.

src/SAML2/Compat/ContainerSingleton.php
tvdijen: Of course it's possible.. Anything can be used as an array index..

> I think it's easier (and clearer) if we specify separately the namespace and the name of the element
Not sure what you mean.. Add two separate entries for one class?

The `qualifiedName` is a naming mistake... Should be localName

jaimeperez: I mean that the method should receive 3 parameters: the fully qualified name of the class we are registering, and the namespace and name of the element it implements.

Actually, that can be simplified if we add a check which I think we need anyway: classes registered should be extending `AbstractXMLElement`. As such, the name of the class should match the name of the element it represents, and it should also have a `NS` constant with its namespace. So we could just pass the class name to `registerClass()`, check that it extends the required `AbstractXMLElement`, and then pick both pieces of information we need from it, and add them as the tuple indexing the registry.

tvdijen:
OK, so the way it is now works as follows:

`registerClass('\This\Is\A\Test');`

Will lead to a registry like this:
```
Array(
    [\This\Is\A\Test] => Array(
            [0] => urn:oasis:names:tc:SAML:2.0:assertion
            [1] => Test
        )
)
```

And then `getRegisteredClass(Constants::NS_SAML, 'Test');` will output:
`\This\Is\A\Test`

jaimeperez:
> OK, so the way it is now works as follows:
> 
> `registerClass('\This\Is\A\Test');`
> 
> Will lead to a registry like this:
> 
> ```
> Array(
>     [\This\Is\A\Test] => Array(
>             [0] => urn:oasis:names:tc:SAML:2.0:assertion
>             [1] => Test
>         )
> )
> ```
> 
> And then `getRegisteredClass(Constants::NS_SAML, 'Test');` will output: `\This\Is\A\Test`

Uhm, shouldn't we index it the other way around? In this case it's a little harder to search for occurrences (while insertion in the registry has constant time in both alternatives), and also this opens up the possibility to have multiple classes registered for the same namespace and element name, leading to ambiguous behaviour.

Also, I just checked and it is in fact not possible to use an array as an index, so we would need to transform that into a string. I'd suggest something like the following:

```php
$key = join(
    ':',
    [
        urlencode($namespace),
        $element
    ]
);
```

That way we can keep both insertions and search at linear complexity (_O(1)_).

PR #225, opened 2020-03-13
Implement encrypted elements

jaimeperez:


src/SAML2/XML/xenc/EncryptedData.php
```diff
@@ -0,0 +1,283 @@
...
+     * @param string|null $id The Id attribute of this object. Optional.
+     * @param string|null $type The Type attribute of this object. Optional.
+     * @param string|null $mimeType The MimeType attribute of this object. Optional.
+     * @param string|null $encoding The Encoding attribute of this object. Optional.
+     * @param EncryptionMethod|null $encryptionMethod The EncryptionMethod object of this EncryptedData. Optional.
+     * @param KeyInfo|null $keyInfo The KeyInfo object of this EncryptedData. Optional.
+     */
+    public function __construct(
+        CipherData $cipherData,
+        ?string $id,
```
tvdijen: Shouldn't all of these nullables be optional?

jaimeperez: Good catch!

tvdijen: Not just here, also in other classes

tvdijen:
One general comment is that we should use fully namespaced class-names in phpdocs

src/SAML2/XML/xenc/EncryptedData.php
jaimeperez: As far as I've seen, `EncryptedKey` as well. Took care of both.

jaimeperez:
> One general comment is that we should use fully namespaced class-names in phpdocs

Ack! I didn't polish everything for now, so most of the code ended up being a bit of a stub autogenerated by PhpStorm.

src/SAML2/EncryptedElementInterface.php
```diff
@@ -0,0 +1,50 @@
+<?php
```
tvdijen, line 1: This file should probably just go into the XML-directory

src/SAML2/EncryptedElementTrait.php
```diff
@@ -0,0 +1,180 @@
+<?php
```
tvdijen, line 1: dito

src/SAML2/XML/IdentifierTrait.php
```diff
@@ -42,7 +42,7 @@ public function getIdentifier(): ?IdentifierInterface
     /**
      * Set the value of the identifier-property
      *
-     * @param \SAML2\XML\saml\IdentifierInterface|null
+     * @param \SAML2\XML\saml\IdentifierInterface|null $identifier
```
tvdijen, line 45: This has been fixed in master already.. Perhaps a rebase against master would be a good idea?

src/SAML2/XML/saml/EncryptedAssertion.php
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * @inheritDoc
+     *
+     * @return \SAML2\XML\saml\Assertion
+     * @throws \Exception
+     *
+     * @psalm-suppress MismatchingDocblockReturnType
+     * @psalm-suppress ImplementedReturnTypeMismatch
+     */
+    public function decrypt(XMLSecurityKey $key, array $blacklist = []): AbstractXMLElement
```
tvdijen, line 37: We may want to avoid the @psalm-suppress by following the same as we did here using the object-return type: https://github.com/simplesamlphp/saml2/blob/6c04f44afd3f8b711cdbb77c964817094bb55359/src/SAML2/XML/AbstractXMLElement.php#L246

src/SAML2/XML/IdentifierTrait.php
jaimeperez: It is!

src/SAML2/XML/saml/EncryptedAssertion.php
jaimeperez: The problem here is that while `fromXML()` returns an object of the same class, `decrypt()` does not. So `self` is not an option, unfortunately (unless I'm misunderstanding your suggestion)...

tvdijen: I never said anything about `self`... I'm talking about using 'object' as a return type and specify the specific type in the phpdoc...

jaimeperez: Using `object` was a mistake, in my opinion. It doesn't give you anything, it doesn't offer any API you are guaranteed to have, even though elements created with `fromXML()` are always supposed to be descendants of `AbstractXMLElement`. At least it should have been the latter.

PR #227, opened 2020-03-17
Add samlp classes

tvdijen:
Implementations for:
- Scoping
- IDPList
- IDPEntry

### DIFF
--- a/src/SAML2/XML/samlp/Scoping.php
+++ b/src/SAML2/XML/samlp/Scoping.php
@@ -1,0 +1,166 @@
+<?php
+
+declare(strict_types=1);
+
+namespace SAML2\XML\samlp;
+
+use DOMElement;
+use SAML2\Constants;
+use SAML2\DOMDocumentFactory;
+use SAML2\Utils;
+use Webmozart\Assert\Assert;
+
+/**
+ * Class for handling SAML2 Scoping.
+ *
+ * @author Tim van Dijen, <tvdijen@gmail.com>
+ * @package simplesamlphp/saml2
+ */
+final class Scoping extends AbstractSamlpElement
+{
+    /** @var \SAML2\XML\samlp\IDPList|null */
+    protected $IDPList;
+
+    /** @var string[] */
+    protected $requesterId;
+
+    /** @var int|null */
+    protected $proxyCount;
+
+
+    /**
+     * Initialize a Scoping element.
+     *
+     * @param int|null $proxyCount
+     * @param \SAML2\XML\samlp\IDPList $idpList
+     * @param string[] $requesterId
+     */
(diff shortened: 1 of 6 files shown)


src/SAML2/XML/samlp/Scoping.php
```diff
@@ -0,0 +1,166 @@
...
+     * @throws \InvalidArgumentException if the qualified name of the supplied element is wrong
+     */
+    public static function fromXML(DOMElement $xml): object
+    {
+        Assert::same($xml->localName, 'Scoping');
+        Assert::same($xml->namespaceURI, Scoping::NS);
+
+        $proxyCount = self::getAttribute($xml, 'ProxyCount', null);
+        $idpList = IDPList::getChildrenOfClass($xml);
+        $requesterId = Utils::extractStrings($xml, AbstractSamlpElement::NS, 'RequesterID');
```
jaimeperez, line 132: This makes me wonder if there is an actual way for `Utils::extractStrings()` to return an array with multiple strings, rather than one single string. Or if it even makes sense.

tvdijen: Not sure what you're saying.. It'll return an array with multiple strings..

jaimeperez: I know, that's the signature, but I wonder if that's actually a legitimate use case. As far as I understand the code in `Utils`, the only way this can happen is when there are multiple elements with text content under the parent. It's just weird in my head.

tvdijen: Yeah, Scoping can contain multiple RequesterIDs.. That's perfectly legit?
In most cases there will be only one result though..

jaimeperez: Uhm, I see! Thanks for the example! It feels kind of weird to handle those cases like this, but nevermind!

tvdijen:
@jaimeperez I've added the nullables-tests.. Could you review the last four commits before I merge this?

tests/SAML2/XML/samlp/IDPListTest.php
```diff
@@ -67,6 +67,35 @@ public function testMarshalling(): void
...
+        $entries = $list->getIdpEntry();
+        $this->assertCount(1, $entries);
+
+        $this->assertEquals('urn:some:requester1', $entries[0]->getProviderID());
+        $this->assertEquals('testName1', $entries[0]->getName());
+        $this->assertEquals('testLoc1', $entries[0]->getLoc());
+
+        $this->assertNull($list->getGetComplete());
+
+        $this->assertEquals($document->saveXML($document->documentElement), strval($list));
```
jaimeperez, line 95: You don't need to load an XML document and then call `saveXML()` on it. You could just compare against the actual string here.

PR #228, opened 2020-03-17
Add saml classes

tvdijen:
Implementations for:
- Condition
- Conditions
- AudienceRestriction
- Audience
- ProxyRestriction
- OneTimeUse
- AbstractConditionType


src/SAML2/XML/saml/Conditions.php
```diff
@@ -0,0 +1,301 @@
...
+        $oneTimeUse = OneTimeUse::getChildrenOfClass($xml);
+        $proxyRestriction = ProxyRestriction::getChildrenOfClass($xml);
+
+        return new self(
+            $notBefore !== null ? Utils::xsDateTimeToTimestamp($notBefore) : null,
+            $notOnOrAfter !== null ? Utils::xsDateTimeToTimestamp($notOnOrAfter) : null,
+            $condition,
+            $audienceRestriction,
+            array_pop($oneTimeUse),
+            array_pop($proxyRestriction)
```
jaimeperez, line 260: I'm thinking about this strategy here... The schema allows for multiple elements, but then the standard explicitly forbids it. Now, the question is: should we just ignore multiple elements and pick the last one (as in this case), or should we raise an error?

@thijskh what's your opinion on this? I believe in other places the strategy has been to raise an error, but now I'm in doubt...

src/SAML2/XML/saml/OneTimeUse.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace SAML2\XML\saml;
+
+use DOMElement;
+use SAML2\Constants;
+use Webmozart\Assert\Assert;
+
+final class OneTimeUse extends AbstractConditionType
```
jaimeperez, line 9: Having a look at the standard, I think we can completely drop this class and use instead a boolean in the classes referencing it. In the end, this element being present signals one thing, and the lack of it signals the opposite. Therefore, in `Conditions` we could just have an `isOneTimeUse()` method as a substitute (maybe even receiving a bool parameter to act both as getter and setter), and that would suffice for the use of this element we need to do. The only downside I see there is that then `Conditions::fromXML()` becomes a bit more complicated, since you need to test for the element, instead of just calling `getChildrenOfClass()`

src/SAML2/XML/saml/ProxyRestriction.php
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * Set the value of the audience-attribute
+     *
+     * @param \SAML2\XML\saml\Audience[] $audience
+     * @return void
+     */
+    protected function setAudience(array $audience): void
+    {
+        $this->audience = $audience;
```
jaimeperez, line 82: A check is needed here as well.

tests/SAML2/CustomCondition.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace SAML2;
+
+use DOMElement;
+use SAML2\XML\saml\Condition;
+use Webmozart\Assert\Assert;
+
+final class CustomCondition extends Condition
```
jaimeperez, line 9: Right now I don't remember the exact approach we took before, but in this case, wouldn't it be enough to do this in `Condition`? Wouldn't the same approach of `BaseID` work here?

src/SAML2/XML/saml/Conditions.php
tvdijen: I think our strategy has been to be less strict on the receiving end and strict on what we generate

tests/SAML2/CustomCondition.php
tvdijen: I think we came to the conclusion that our BaseID implementation was wrong, because it doesn't consider any child-elements and/or XML-attributes... Not sure though

src/SAML2/XML/saml/OneTimeUse.php
tvdijen: I went for consistency here, but I can change it if you like

tests/SAML2/CustomCondition.php
jaimeperez: Uhm, isn't it possible in those cases to just extend `BaseID` to implement whatever you need as child elements/attributes?

src/SAML2/XML/saml/OneTimeUse.php
jaimeperez: I think it would be simpler to use, because the way it is now you need to call the getter, and see if it's not `null`, which in my head is a very convoluted way to test a boolean condition. Also, having this as an independent class, means we need to test it as well...

src/SAML2/XML/saml/Conditions.php
jaimeperez: Yeah, I remember discussing about that, and it definitely makes sense!

tests/SAML2/CustomCondition.php
tvdijen: I don't recall exactly.. Anyway, for BaseID we also created a CustomID test-case..

PR #230, opened 2020-04-23
Rewrite assertion

### LINKED ISSUE #192: Add static fromXML-methods to all classes in the SAML2\XML namespace
We would basically need to copy the contents of the constructor to a static fromXML-method.
Next step is to change to API on these classes to be able to set all values from the constructor at once.

tvdijen:
Closes #192


src/SAML2/XML/saml/SubjectLocality.php
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * Set the value of the address-property
+     *
+     * @param string|null $address
+     * @return void
+     */
+    private function setAddress(?string $address): void
+    {
+        $this->address = $address;
```
jaimeperez, line 60:
```php
        Assert::nullOrIp($address, 'Invalid IP address');
```

src/SAML2/XML/saml/SubjectLocality.php
```diff
@@ -0,0 +1,140 @@
...
+
+
+    /**
+     * Set the value of the dnsName-property
+     *
+     * @param string|null $dnsName
+     * @return void
+     */
+    private function setDnsName(?string $dnsName): void
+    {
```
jaimeperez, line 82:
```php
        Assert::nullOrStringNotEmpty($dnsName, 'Invalid DNS name');
```

src/SAML2/XML/saml/SubjectLocality.php
tvdijen: I think we've had this discussion before in another place, but the schema says it's a `string`

jaimeperez: Yes, it is, but a string that represents an IP address. Maybe I'm misunderstanding `nullOrIp()`, but that's what I would expect it to test for.

tvdijen: Don't you want to be liberal on the incoming end?

I remember adding similar restrictions on one of the mdui or mdrpi classes (v4) and then I had to undo it because people started to complain

tvdijen: Here it is;  https://github.com/simplesamlphp/saml2/commit/1e67636360d7062b4e677cea9602a56198aceba3

jaimeperez: Yeah, in most occasions we should be flexible, but in others I'm not sure how much sense does it make to be so... In this case, the address is supposed to be exactly an IP address, so I don't really see the point of allowing anything else. What we could do in cases like this, though, is to continue but discarding the value.

tvdijen: The other case was a bit more vague and described the value as 'network location'.. I think someone had a hostname in there...  I also remember a case of mdui iconUrl where the value was a base64-encoded image instead of a url..

Let's leave it for now

jaimeperez: Yes, those cases make full sense. One thing is validating input, and another different thing is discarding input that's actually legitimate. But in this case I think it should be fine to check for an IP.

In any case, not a big deal.

src/SAML2/XML/saml/AttributeStatement.php
```diff
@@ -0,0 +1,133 @@
...
+
+    /** @var \SAML2\XML\saml\EncryptedAttribute[] */
+    protected $encryptedAttributes = [];
+
+
+    /**
+     * AttributeStatement constructor.
+     *
+     * @param \SAML2\XML\saml\Attribute[]|null $attributes
+     * @param \SAML2\XML\saml\EncryptedAttribute[]|null $encryptedAttributes
```
tvdijen, line 29: |null is wrong here (twice)

src/SAML2/XML/saml/AttributeStatement.php
```diff
@@ -0,0 +1,133 @@
...
+     * @return \SAML2\XML\saml\EncryptedAttribute[]
+     */
+    public function getEncryptedAttributes(): array
+    {
+        return $this->encryptedAttributes;
+    }
+
+
+    /**
+     * @param XMLSecurityKey $key
```
tvdijen, line 70: Full name here

src/SAML2/XML/saml/AttributeStatement.php
```diff
@@ -0,0 +1,133 @@
...
+     * @param array $blacklist
+     *
+     * @throws \Exception
+     */
+    public function decryptAttributes(XMLSecurityKey $key, array $blacklist = []): void
+    {
```
tvdijen, line 76: Assert::allStringNotEmpty($blacklist);

src/SAML2/XML/saml/AttributeStatement.php
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * @param DOMElement $xml
+     *
+     * @return object
+     */
+    public static function fromXML(DOMElement $xml): object
+    {
+        Assert::same($xml->localName, 'AttributeStatement');
+        Assert::same($xml->namespaceURI, AttributeStatement::NS);
```
tvdijen, line 110: Add InvalidDOMElement as parameter (twice)

src/SAML2/XML/saml/AuthnStatement.php
```diff
@@ -179,8 +176,10 @@ private function setSubjectLocality(?SubjectLocality $subjectLocality): void
      * Convert XML into a AuthnContext
      *
      * @param \DOMElement $xml The XML element we should load
+     *
      * @return \SAML2\XML\saml\AuthnContext
-     * @throws \InvalidArgumentException if the qualified name of the supplied element is wrong
+     * @throws \InvalidArgumentException if the qualified name of the supplied element is wrong*@throws \Exception
```
tvdijen, line 181: Remove last part

src/SAML2/XML/saml/Statement.php
```diff
@@ -0,0 +1,104 @@
...
+use Webmozart\Assert\Assert;
+
+/**
+ * Class implementing the <saml:Statement> extension point.
+ *
+ * @package simplesamlphp/saml2
+ */
+class Statement extends AbstractStatement
+{
+
```
tvdijen, line 18: Remove newline

src/SAML2/XML/saml/Statement.php
```diff
@@ -0,0 +1,104 @@
...
+     * Convert XML into an Statement
+     *
+     * @param \DOMElement $xml The XML element we should load
+     *
+     * @return \SAML2\XML\saml\Statement
+     * @throws \InvalidArgumentException  If xsi:type is not defined or does not implement IdentifierInterface
+     */
+    public static function fromXML(DOMElement $xml): object
+    {
+        Assert::same($xml->localName, 'Statement');
```
tvdijen, line 78: Add InvalidDOMElement (twice)

tvdijen: The notNull seems overkill?

src/SAML2/XML/saml/Assertion.php
```diff
@@ -275,179 +186,165 @@ public function getConditions(): ?Conditions
...
     /**
-     * @param \DOMNode $attribute
-     * @param string   $attributeName
-     * @return void
+     * Set the statements in this assertion
+     *
+     * @param \SAML2\XML\saml\AbstractStatement[] $statements
      */
-    private function parseAttributeValue(DOMNode $attribute, string $attributeName): void
+    protected function setStatements(array $statements): void
```
tvdijen, line 238: Needs an assertion to test the array values

src/SAML2/XML/saml/Assertion.php
```diff
@@ -457,23 +354,24 @@ private function parseSignature(DOMElement $xml): void
      * Otherwise, true will be returned. An exception is thrown if the
      * signature validation fails.
      *
-     * @param  XMLSecurityKey $key The key we should check against.
+     * @param XMLSecurityKey $key The key we should check against.
```
tvdijen, line 357: Full class name

src/SAML2/XML/saml/Assertion.php
```diff
@@ -186,471 +84,274 @@ class Assertion implements SignedElementInterface
...
-            $authnContextEl,
-            Constants::NS_SAML,
-            'AuthenticatingAuthority'
-        );
+    /**
+     * @return \SAML2\XML\saml\AuthnStatement[]
+     */
+    public function getAuthnStatements(): array
+    {
+        return array_filter($this->statements, function ($statement) {
```
tvdijen, line 221: This unfortunately doesn't work for mock...

src/SAML2/Assertion/Transformer/DecodeBase64Transformer.php
```diff
@@ -44,15 +47,40 @@ public function transform(Assertion $assertion): Assertion
...
+                }
+                $decodedAttributes[] = new Attribute(
+                    $attribute->getName(),
+                    $attribute->getNameFormat(),
+                    $attribute->getFriendlyName(),
+                    $values,
+                    $attribute->getAttributesNS()
+                );
             }
+            $statements[] = new AttributeStatement($decodedAttributes);
```
tvdijen, line 70: We still have to do something here for the EncryptedAttributes..

src/SAML2/Assertion/Validation/ConstraintValidator/SessionNotOnOrAfter.php
```diff
@@ -19,7 +19,7 @@ class SessionNotOnOrAfter implements
      */
     public function validate(Assertion $assertion, Result $result): void
     {
-        $sessionNotOnOrAfterTimestamp = $assertion->getSessionNotOnOrAfter();
+        $sessionNotOnOrAfterTimestamp = $assertion->getAuthnStatements()[0]->getSessionNotOnOrAfter();
```
tvdijen, line 22: We have to consider multiple AuthnStatements

src/SAML2/XML/saml/AttributeStatement.php
```diff
@@ -85,6 +85,7 @@ public function decryptAttributes(XMLSecurityKey $key, array $blacklist = []): v
 
     /**
      * @return bool
+     * @return void
```
MKodde, line 88: Added this return type by accident?

MKodde (commented):
I must confess, this was a touch cookie. The SAML spec is not imprinted in my long term memory. So I mainly looked at this from a code review perspective. And I must say, it all looked solid. See two comments, but overall things look very good. Love it when more than half of the change set consist of tests!

src/SAML2/XML/saml/AttributeStatement.php
tvdijen: Must have been a rebase mistake.. Taking it out!

src/SAML2/Assertion/Transformer/DecodeBase64Transformer.php
```diff
@@ -43,17 +46,44 @@ public function transform(Assertion $assertion): Assertion
...
+                $values = [];
+                foreach ($attribute->getAttributeValues() as $encodedValue) {
+                    $encoded = $encodedValue->getValue();
+                    if (is_string($encoded)) {
+                        foreach ($this->decodeValue($encoded) as $decoded) {
+                            $values[] = new AttributeValue($decoded);
+                        }
+                    } else {
+                        $values[] = $encodedValue;
+                    }
```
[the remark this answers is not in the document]
tvdijen: It was kind of a case where I had to make existing code working with the new XML-classes.. I guess I could try and move some of it to private methods

tvdijen:
Thanks for your review Michiel! We put a lot of effort in getting near-perfect coverage for the XML-directory, taking the coverage of the entire library from 60 to around 85%.. That should pay of in the future.

src/SAML2/Assertion/Transformer/DecodeBase64Transformer.php
```diff
@@ -87,6 +77,23 @@ public function transform(Assertion $assertion): Assertion
...
+            if (is_string($v)) {
+                foreach ($this->decodeValue($v) as $decoded) {
+                    $v[$k] = new AttributeValue($decoded);
+                }
+            }
+        });
+        return $encodedValues;
+    }
+
+
```
tvdijen, line 96: @MKodde I'm a bit rusty and I can't seem to get this 'fix' to work..  What am I missing here?

tvdijen: Sorry, Jaime and I spent all night on this and we were left frustrated.. I rolled back eventually

PR #231, opened 2020-05-22
Fix example not using Issuer object

webmaster777:
This should be back-ported/cherrypicked on 4.x branch as well.


tvdijen:
This is not how it should be done on master.. the api is significantly different, so this should be 4.x only

webmaster777:
Wait, the master is behind on the 4.x branch?

tvdijen:
No, what I'm trying to say is that master is way ahead and that the API has changed completely..  No more setter-methods, everything has to go through the constructor (i.e. immutable objects).

I'll change this PR to fix the readme for 4.x and write another one for master..

tvdijen:
Changing the base-branch didn't work out very well, so I ended up cherry-picking your commit onto the 4.x-branch and I've committed a similar fix to the master-branch.. Thanks for reporting!!

PR #232, opened 2020-06-09
Support Request Initiator metadata extension

tvdijen:


src/SAML2/XML/init/RequestInitiator.php
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Set the value of the Binding property.
+     *
+     * @param string $binding
+     * @throws \InvalidArgumentException if the Binding is empty
+     */
+    protected function setBinding(string $binding): void
+    {
+        Assert::notEmpty($binding, 'The Binding of an endpoint cannot be empty.');
+        Assert::eq(
```
tvdijen, line 37: This comparison should probably not be case-sensitive

jaimeperez: Uhm, not sure, actually. I think XML namespaces are case-sensitive, right?

src/SAML2/XML/md/Extensions.php
```diff
@@ -73,6 +74,9 @@ public static function fromXML(DOMElement $xml): object
                 'DigestMethod'  => DigestMethod::class,
                 'SigningMethod' => SigningMethod::class,
             ],
+            RequestInitiator::NS             => [
```
jaimeperez, line 77: Nitpick: a bit extra space here?

src/SAML2/XML/init/RequestInitiator.php
```diff
@@ -0,0 +1,45 @@
...
+    public const NS_PREFIX = 'init';
+
+
+    /**
+     * Set the value of the Binding property.
+     *
+     * @param string $binding
+     * @throws \InvalidArgumentException if the Binding is empty
+     */
+    protected function setBinding(string $binding): void
```
jaimeperez, line 34: Considering the binding is fixed, does it make sense having a `setBinding()` method?

src/SAML2/XML/init/RequestInitiator.php
tvdijen: You're right, I dunno what I was thinking

src/SAML2/XML/init/RequestInitiator.php
tvdijen: Well the alternative is overriding the constructor.. I thought it would make sense to keep the constructor the same as for other AbstractEndpoints

PR #239, opened 2020-08-15
Implement new CertificatesMock from test-framework

tvdijen:


tvdijen:
We need an signedassertion.xml signed by the cert from the test-framework for the tests to pass

PR #240, opened 2020-08-18
Make processor aware of assertion types

MKodde:
Both the Encrypted and regular Assertion classes should be passable in the decryptAssertion method. If not, you would never be able to process a SAML Response consisting of regular Assertion objects.

The test merely verifies the behavior that was changed in this commit. No additional processor tests where added. But that should be simple enough in the future.

### DIFF
--- a/src/SAML2/Assertion/Processor.php
+++ b/src/SAML2/Assertion/Processor.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace SAML2\Assertion;
 
 use Psr\Log\LoggerInterface;
-
 use SAML2\Assertion;
 use SAML2\Assertion\Exception\InvalidAssertionException;
 use SAML2\Assertion\Exception\InvalidSubjectConfirmationException;
@@ -15,7 +14,6 @@ use SAML2\Assertion\Validation\SubjectConfirmationValidator;
 use SAML2\Configuration\IdentityProvider;
 use SAML2\EncryptedAssertion;
 use SAML2\Response\Exception\InvalidSignatureException;
-use SAML2\Response\Exception\UnencryptedAssertionFoundException;
 use SAML2\Signature\Validator;
 use SAML2\Utilities\ArrayCollection;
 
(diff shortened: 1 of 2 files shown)


src/SAML2/Assertion/Processor.php
```diff
@@ -95,7 +95,13 @@ public function decryptAssertions(ArrayCollection $assertions)
     {
         $decrypted = new ArrayCollection();
         foreach ($assertions->getIterator() as $assertion) {
-            $decrypted->add($this->decryptAssertion($assertion));
+            if ($assertion instanceof EncryptedAssertion) {
+                $decrypted->add($this->decryptAssertion($assertion));
+            } elseif ($assertion instanceof Assertion) {
+                $decrypted->add($assertion);
+            } else {
+                throw new InvalidAssertionException('The assertion is neither of type: EncryptedAssertion nor Assertion.');
```
pmeulen, line 103: Nitpick: using neither ... nor over a colon looks weird. 
Suggestions:
"The assertion is neither of type EncryptedAssertion nor of type Assertion."
"The assertion must be of type: EncryptedAssertion or Assertion"

thijskh (commented):
It looks correct to me, and makes the method behave conforming to how the docblock that describes that it will. But would prefer if Tim or Jaime could comment whether this is the way to go.

It should also be merged to master so it will not break in 5.x again.

MKodde:
Thanks for the review @pmeulen and @thijskh! Good suggestions.

MKodde:
@tvdijen the unit tests seem somewhat fragile. Autoload problems occurred after adding the ProcessorTest. Only, they did not occur on my tests but on some other test cases. They where caused by me using a data provider. Removing the provider and using the `runTestsInSeparateProcesses` annotation, the problem is fixed. 

Is this something you are aware of? Or am I doing something wrong here?

tvdijen:
I haven't really paid much attention to this part of the library.. Jaime and I mostly worked on the classes in the XML-dir..
The validators/processors are a bit hard to follow, but this change seems fine.
The issue iwth the dataprovider surprises me.. I'm pretty sure we use them in other places.. We should take a closer look at  the tests for v5 and try to avoid the use of mocks whenever possible..

tvdijen (approved):

MKodde:
Hi Tim, on our OpenConext phpunit tests we never have issues with Mockery mocks. Maybe a side by side comparison of the PHPUnit/PSR autoloading will give some insights. Although i did not find a problem in plain sight yesterday.

Regarding hoisting this change to master; what would you like me to do? Open a new PR or simply cherry pick these commits on master?

tvdijen:
Cherry-picking is fine, just make sure to use the correct namespaces for the Assertion/EncryptedAssertion classes.. They have been changed.

MKodde:
Opened #242 to get these changes into master

PR #242, opened 2020-08-19
Make processor aware of assertion types

MKodde:
Integrates #240  into `master`

### DIFF
--- a/src/SAML2/Assertion/Processor.php
+++ b/src/SAML2/Assertion/Processor.php
@@ -93,7 +93,13 @@ class Processor
     {
         $decrypted = new ArrayCollection();
         foreach ($assertions->getIterator() as $assertion) {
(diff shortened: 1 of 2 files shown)


tvdijen (approved):

tvdijen:
It's a nitpick, but could you add a bunch of use-statements to the test-file and squash the full namespaced class names?

MKodde:
Absolutely. I thought the team preferred having less import statements in favor of instant readability what namespace a class hails from?

tvdijen:
I've noticed that some of the analyzer tools do a better job with the use-statements in place..
We add the full namespaced classes for parameters to the phpdoc and i'm considering adding @uses tags for the others for instant readability..
It's hard to keep psr12 compliant if you put then inline in the code

PR #254, opened 2020-12-03
Support ECP from Microsoft

melanger:
Add content type application/soap+xml for SOAP requests (ECP).

In our setup, Microsoft sends ECP requests with Content-type: "application/soap+xml". SimpleSAMLphp does not recognize them as SOAP requests.

There was actually a commit e5bfe333eb3bbcf295d13215ff7e9594181c642e for PR #48 which added this, but the PR was closed an replaced by #96, which did not include this content type.


tvdijen (approved):

tvdijen:
Thanks @melanger ! makes sense! I've tagged a new release and will make sure it ends up in SSP 1.19

PR #260, opened 2021-03-16
Add LocalizedName elements for md & mdui namespace

tvdijen:
These classes were never implemented, but they're needed if we want to add toArray/fromArray for these classes

### DIFF
--- a/src/SAML2/XML/mdui/UIInfo.php
+++ b/src/SAML2/XML/mdui/UIInfo.php
@@ -42,9 +42,9 @@ final class UIInfo extends AbstractMduiElement
     protected array $Description = [];
 
     /**
-     * The InformationURL, as an array of language => url.
+     * The InformationURL, as an array of InformationURL objects
      *
-     * @var string[]
+     * @var \SimpleSAML\SAML2\XML\mdui\InformationURL[]
      */
     protected array $InformationURL = [];
 
@@ -75,7 +75,7 @@ final class UIInfo extends AbstractMduiElement
      *
      * @param \SimpleSAML\SAML2\XML\mdui\DisplayName[] $DisplayName
      * @param \SimpleSAML\SAML2\XML\mdui\Description[] $Description
-     * @param string[] $InformationURL
+     * @param \SimpleSAML\SAML2\XML\mdui\InformationURL[] $InformationURL
      * @param string[] $PrivacyStatementURL
      * @param \SimpleSAML\SAML2\XML\mdui\Keywords[] $Keywords
      * @param \SimpleSAML\SAML2\XML\mdui\Logo[] $Logo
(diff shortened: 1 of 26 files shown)


src/SAML2/XML/mdui/UIInfo.php
```diff
@@ -334,16 +334,24 @@ public static function fromXML(DOMElement $xml): object
...
+                        break;
+                    case 'Description':
+                        $Description[] = Description::fromXML($node);
+                        break;
+                    case 'InformationURL':
+                        $InformationURL[] = InformationURL::fromXML($node);
+                        break;
+                    case 'PrivacyStatementURL':
+                        $PrivacyStatementURL[] = PrivacyStatementURL::fromXML($node);
+                        break;
```
jaimeperez, line 354: Couldn't we here use `getChildrenOfClass()`?

tests/SAML2/XML/mdui/UIInfoTest.php
```diff
@@ -215,15 +231,15 @@ public function testUnmarshalling(): void
...
-        $this->assertEquals('http://www.example.edu/en/', $uiinfo->getInformationURL()['en']);
-        $this->assertEquals('http://www.example.edu/', $uiinfo->getInformationURL()['el']);
+        $this->assertEquals('http://www.example.edu/en/', $uiinfo->getInformationURL()[0]->getValue());
+        $this->assertEquals('http://www.example.edu/', $uiinfo->getInformationURL()[1]->getValue());
         $this->assertCount(1, $uiinfo->getPrivacyStatementURL());
-        $this->assertEquals('https://example.org/privacy', $uiinfo->getPrivacyStatementURL()['en']);
+        $this->assertEquals('https://example.org/privacy', $uiinfo->getPrivacyStatementURL()[0]->getValue());
         $this->assertCount(1, $uiinfo->getDescription());
-        $this->assertEquals('Just an example', $uiinfo->getDescription()['en']);
+        $this->assertEquals('Just an example', $uiinfo->getDescription()[0]->getValue());
```
jaimeperez, line 242: Should we also check the language tag for each object here?

jaimeperez (approved):

src/SAML2/XML/mdui/UIInfo.php
tvdijen: We could rewrite this entire method to:
- Use getChildrenOfClass for everything we need and;
- Use the for-loop to catch everything else

Ideally we'd want to avoid creating Chunks for classes we know, but that would require our class registry in place

tests/SAML2/XML/mdui/UIInfoTest.php
tvdijen: Probably, yes!  I'l take care of it

PR #266, opened 2021-04-05
Fix Condition structure

tvdijen:


src/SAML2/Compat/AbstractContainer.php
```diff
@@ -92,6 +97,8 @@ public function registerExtensionHandler(string $class): void
         Assert::subclassOf($class, AbstractXMLElement::class);
         if (is_subclass_of($class, CustomIdentifierInterface::class, true)) {
             $key = $class::getXsiType() . ':BaseID';
+        } elseif (is_subclass_of($class, Condition::class, true)) {
+            $key = $class::getXsiType();
```
jaimeperez, line 101: Shouldn't this be the same as the case for `BaseID`?

tvdijen: Not if we put ssp:CustomCondition in the class constant....  It's something we have to find out at some point that works best

PR #268, opened 2021-05-07
Rewrite Condition #2

tvdijen:


tvdijen:
@jaimeperez  Trying to peel of where I went bad, so this is hopefully back to the part where at #266 you made your comment..

PR #276, opened 2021-05-07
Add & implement md:EmailAddress & md:TelephoneNumber

tvdijen:


tvdijen:
@jaimeperez This one may be debatable.. We've had discussions before about our classes representing actual XML, but EMailAddress has the 'mailto' quirk.. We don't want to be too strict and refuse emails without the prefix..   I think this is (so far) the only exception to the rule and I think I did solve it in a clever way, but please take a close look!

PR #281, opened 2021-08-01
trim decrypted xml to avoid exception

phelan:
when the assertion xml is start with a "\n",  $newDoc->firstChild->firstChild got a DOMText but not DOMElement, trim xml will fix this

### DIFF
--- a/src/SAML2/Utils.php
+++ b/src/SAML2/Utils.php
@@ -496,7 +496,7 @@ class Utils
          */
         $xml = '<root xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" '.
                         'xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">'.
-            $decrypted.
+            trim($decrypted).
             '</root>';
 
         try {


tvdijen:
Thanks @phelan ! Makes total sense!  I've asked my fellow dev to look at this before we merge it

tvdijen:
@phelan How did you end up with an assertion that starts with a newline? Was it created using this library, or is it some other product send it to you?
We kinda feel that the newline should have never ended up in the encrypted assertion in the first place.

phelan:
A customer's idp send it to me, i can't ask them to trim the assertion, only have to trim it myself

tvdijen:
Can you ask them what software they're using for their IDP? Or can you perhaps derive this info from their metadata?

phelan:
@tvdijen  I find this in metadata, may be SAML Artifact Resolution Services
<ns4:ArtifactResolutionService index="0" isDefault="false" Binding="urn:oasis:names:tc:SAML:2.0:bindings:SOAP"/>

tvdijen:
That's not a product..  That's just part of the SAML protocol..   Sometimes the URLs in the metadata give away the name of the product.

phelan:
I cant't find any other useful information in the metadata, may be they write the wrong code to build the assertion xml, because a well-formed XML documents can't starts with a newline character, i will checkout with them, thanks @tvdijen

PR #282, opened 2021-08-06
Added option to set KeyName for KeyDescriptor

tommy2d:
Since some IDP parties seem to require the Keyname to be set in the xml metada, I added the keyName argument to the createKeyDescriptor method.

### DIFF
--- a/src/SAML2/Utils.php
+++ b/src/SAML2/Utils.php
@@ -20,6 +20,7 @@ use SAML2\Exception\RuntimeException;
 use SAML2\XML\ds\KeyInfo;
 use SAML2\XML\ds\X509Certificate;
 use SAML2\XML\ds\X509Data;
+use SAML2\XML\ds\KeyName;
 use SAML2\XML\md\KeyDescriptor;
 
 /**
@@ -649,23 +650,30 @@ class Utils
     /**
      * Create a KeyDescriptor with the given certificate.
      *
-     * @param string $x509Data The certificate, as a base64-encoded DER data.
+     * @param string|null $x509Data The certificate, as a base64-encoded DER data.
+     * @param string|null $keyName The name of the key as specified in the KeyInfo
      * @return \SAML2\XML\md\KeyDescriptor The keydescriptor.
      */
-    public static function createKeyDescriptor(string $x509Data) : KeyDescriptor
+    public static function createKeyDescriptor(string $x509Data = null, string $keyName = null) : KeyDescriptor
     {
-        $x509Certificate = new X509Certificate();
-        $x509Certificate->setCertificate($x509Data);
-
-        $x509Data = new X509Data();
-        $x509Data->addData($x509Certificate);
-
         $keyInfo = new KeyInfo();
-        $keyInfo->addInfo($x509Data);
+
+        if ($keyName !== null) {
+            $keynameEl = new KeyName();
+            $keynameEl->setName($keyName);
+            $keyInfo->addInfo($keynameEl);
+        }
+
+        if ($x509Data !== null) {
+            $x509Certificate = new X509Certificate();
+            $x509Certificate->setCertificate($x509Data);
+            $x509Data = new X509Data();
+            $x509Data->addData($x509Certificate);
+            $keyInfo->addInfo($x509Data);
+        }
 
         $keyDescriptor = new KeyDescriptor();
         $keyDescriptor->setKeyInfo($keyInfo);
-
         return $keyDescriptor;
     }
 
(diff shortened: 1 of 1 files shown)


tvdijen:
Thanks @tommy2d ! We had already fixed this in our rewrite of this library in `master`, but I see no harm in backporting it.. Just out of curiousity, what IDP software requires this? I've seen a lot, but this is a first!

tvdijen:
Just for reference, the xsd-schema is as follows;

```
<complexType name="KeyInfoType" mixed="true">
  <choice maxOccurs="unbounded">     
    <element ref="ds:KeyName"/> 
    <element ref="ds:KeyValue"/> 
    <element ref="ds:RetrievalMethod"/> 
    <element ref="ds:X509Data"/> 
    <element ref="ds:PGPData"/> 
    <element ref="ds:SPKIData"/>
    <element ref="ds:MgmtData"/>
  </choice>
  <attribute name="Id" type="ID" use="optional"/>
</complexType>
```

Meaning ds:KeyInfo must contain at least one of the elements mentioned.. Doesn't matter which one, can also contain more (if not all)

src/SAML2/Utils.php
```diff
@@ -649,23 +650,30 @@ public static function addStrings(
     /**
      * Create a KeyDescriptor with the given certificate.
      *
-     * @param string $x509Data The certificate, as a base64-encoded DER data.
+     * @param string|null $x509Data The certificate, as a base64-encoded DER data.
+     * @param string|null $keyName The name of the key as specified in the KeyInfo
      * @return \SAML2\XML\md\KeyDescriptor The keydescriptor.
      */
-    public static function createKeyDescriptor(string $x509Data) : KeyDescriptor
+    public static function createKeyDescriptor(string $x509Data = null, string $keyName = null) : KeyDescriptor
```
jaimeperez, line 657: The way the signature of the function is right now, it doesn't allow the use case supported by the code of a `$keyName` being passed but no `$x509Data`, since it's typed and not nullable. So this signature allows only for three cases:

* Passing the certificate and the key name.
* Passing the certificate but no key name (backwards-compatible).
* Not passing any of them at all.

I suspect the one that's missing (passing only key name) is also desired looking at the code, so we should change the method's signature to make `$x509Data` nullable (probably both arguments, actually). If my assumption is wrong, then I guess we should change the code to make sure we always include a certificate (and the signature to remove the default value from `$x509Data`).

tommy2d:
This change, and everything I'm about to push for simplesaml itself, is necessary for TVS ('Toegangsverleningsservice'), an authentication platform that is actively being pushed by the Dutch government. One of its famous use-cases is the CoronaCheck app.

src/SAML2/Utils.php
```diff
@@ -649,23 +650,30 @@ public static function addStrings(
...
+        if ($x509Data !== null) {
+            $x509Certificate = new X509Certificate();
+            $x509Certificate->setCertificate($x509Data);
+            $x509Data = new X509Data();
+            $x509Data->addData($x509Certificate);
+            $keyInfo->addInfo($x509Data);
+        }
 
         $keyDescriptor = new KeyDescriptor();
         $keyDescriptor->setKeyInfo($keyInfo);
```
jaimeperez, line 676: The old code expected a value in `$x509Data`. However, the new implementation allows for both `$x509Data` and `$keyInfo` to be missing, and that would imply an empty `KeyInfo` element being added here. According to the schema, `KeyInfo` must have at least one child element, so our code could potentially allow for building non schema-compliant SAML documents.

I guess the new implementation should enforce that one value or the other are present, but never both absent at the same time.

jaimeperez (changes requested):

src/SAML2/Utils.php
tommy2d: I agree. I added a check and a throw.

src/SAML2/Utils.php
tommy2d: I added nullable types to the signature.

tvdijen:
Ah, you are a contractor for DICTU?  I'm working for SSC-ICT ;)

jaimeperez (approved):

tvdijen:
I've tagged v4.2.2 for you @tommy2d !   I'd be interested to hear how this library and SSP is being used by TVS. You think you have a spare moment to fill me in?

tommy2d:
> I've tagged v4.2.2 for you @tommy2d ! I'd be interested to hear how this library and SSP is being used by TVS. You think you have a spare moment to fill me in?

Please send me an email so we can schedule a Google meet or something similar.

PR #284, opened 2021-08-24
Call Utils/HTTP non-statically.

thijskh:
Fixes fatal error in SSP master when starting authentication:
"Using $this when not in object context in lib/SimpleSAML/Utils/HTTP.php:985"

### DIFF
--- a/src/SAML2/Compat/Ssp/Container.php
+++ b/src/SAML2/Compat/Ssp/Container.php
@@ -19,6 +19,25 @@ class Container extends AbstractContainer
      */
     protected $logger;
 
+    /**
+     * @var \SimpleSAML\Utils\HTTP
+     */
+    protected $utilsHttp;
+
+    /**
+     * @var \SimpleSAML\Utils\Random
+     */
+    protected $utilsRandom;
+
+    /**
+     * @var \SimpleSAML\Utils\System
+     */
+    protected $utilsSystem;
+
+    /**
+     * @var \SimpleSAML\Utils\XML
+     */
+    protected $utilsXml;
 
     /**
      * Create a new SimpleSAMLphp compatible container.
@@ -26,6 +45,10 @@ class Container extends AbstractContainer
     public function __construct()
     {
         $this->logger = new Logger();
+        $this->utilsHttp = new HTTP();
+        $this->utilsRandom = new Random();
+        $this->utilsSystem = new System();
+        $this->utilsXml = new XML();
     }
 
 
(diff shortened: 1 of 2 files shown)


tvdijen (approved):

thijskh:
Looking at the rest of the file, seems we need to fix XML, System and Random in the same way?

tvdijen:
Yup, and there is at least one more reference to the HTTP-utils in `src/SAML2/HTTPArtifact.php`..

src/SAML2/HTTPArtifact.php
```diff
@@ -66,8 +66,8 @@ public function getRedirectURL(Message $message) : string
         if ($destination === null) {
             throw new \Exception('Cannot get redirect URL, no destination set in the message.');
         }
-        /** @psalm-suppress UndefinedClass */
-        return HTTP::addURLparameters($destination, $params);
+        $utilsHttp = new HTTP();
```
tvdijen, line 69: Should we use the one from the container instead of instantiating a new one?

thijskh: Not sure it's worth the extra interface and code for one instance across the codebase?

src/SAML2/Compat/Ssp/Container.php
```diff
@@ -19,13 +19,36 @@ class Container extends AbstractContainer
      */
     protected $logger;
 
+    /**
+     * @var \SimpleSAML\Utils\HTTP
+     */
+    protected $utilsHttp;
+
+    /**
+     * @var \SimpleSAML\Utils\Rancom
```
tvdijen, line 28: typo

src/SAML2/HTTPArtifact.php
tvdijen: The thing is, HTTP-Artifact can now only be used together with SSP, not just with the standalone library.
Using the container allows for DI

thijskh: Not sure how to do this in a way that keeps the container interface generic. Maybe solve that separately from this buffix PR.

tvdijen: Oh but wait, we moved the SSPContainer to SSP because it didn't belong here..  See SSP@master

thijskh: So this fix is ok for 4.x I presume.

tvdijen: Yeah, let's leave it as is.. I don't see a way to properly deal with this without breaking backwards compatibility, so it'll have to wait for 5.x

PR #285, opened 2021-09-07
SAML Exceptions

tvdijen:
I've imported the classes from SimpleSAMLphp and added the ones that were missing..
I've removed the `Throwable $cause` parameter because it wasn't used anywhere.

The original `SimpleSAML\Module\saml\Error` used to extend `SimpleSAML\Error\Error`, but not extends `RuntimeException`.. That shouldn't matter, because it's still converted to an SSP error [here](https://github.com/simplesamlphp/simplesamlphp/blob/master/www/_include.php#L16)

The goal here is to remove these classes from SSP (the saml-module should probably go away).  Tests failed before this PR


tvdijen:
We could easily backport this into 4.x and remove these classes from the SSP saml-module in 2.0

tvdijen:
One more thing  I've been discussing with Jaime is that we could split them into two branches;  Responder-exceptions and Requester-exceptions.. Downside;  some exceptions can occur on both responder/requester end, so we'd have to create two classes for them..  The upside would be that we can drop the `$responsible` parameter, making them usable in `Assert`s

tvdijen:
So, I think I got of on the wrong foot here.. It is not always possible to know whether an error is caused by the requester or the responder. We should throw an exception and let the catcher decide what it is.
This also means that `AbstractProtocolException` should probably go back to SimpleSAMLphp instead of into this library.

PR #289, opened 2022-01-05
Throw a specific exception for unsupported binding.

thijskh:
Either whether there's a type specified that we don't know or if the guessing failed (the exception message will discern the two cases). This allows SimpleSAMLphp to handle this more neatly.

### DIFF
--- a/src/SAML2/Binding.php
+++ b/src/SAML2/Binding.php
@@ -6,7 +6,7 @@ namespace SimpleSAML\SAML2;
 
 use Exception;
 use SimpleSAML\SAML2\XML\samlp\AbstractMessage;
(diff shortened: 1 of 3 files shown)


tvdijen:
Duplicate of #285

thijskh:
It's not that obvious to me how inside getBinding/getCurrentBinding we would decide if it's the requester or responder's "fault"?

tvdijen:
You're right, and it's probably irrelevant.. The exception should be caught somewhere where it _is_ possible to determine if it's a requester/responder issue..

tvdijen (approved):

tvdijen:
I suspect this is something you need for OC?  I will backport this to a new 4.5 branch.. We may be able to use some of this in SSP 2.0

Issue #295, opened 2022-05-23
PHP 8.1 fixes in AttributeValue are broken


tommy2d:
The __serialize and __unserialize additions in AttributeValue are causing fatal errors when simplesaml is used with the Dutch Digid IDP:

: PHP Fatal error:  Uncaught TypeError: SAML2\XML\saml\AttributeValue::getElement(): Return value must be of type DOMElement, null returned

I can work around this issue by commenting out both the __serialize() and __unserialize() implementations, falling back to default PHP behavior. 

How to proceed? does it help to provide serialized session contents for both the patched and the unpatched situation?

tvdijen:
My best guess by looking at this is that it breaks when AtributeValue contains an XML-structure instead of SimpleContent.
The way they changed serialization in PHP between 7.4 and 8.1 is absolutely horrible and I'm not sure if we can fix this in a way we can support all those versions at the same time.

tvdijen:
The scenario's are covered by a unit test though: https://github.com/simplesamlphp/saml2/blob/release-4.x/tests/SAML2/XML/saml/AttributeValueTest.php#L94-L118

tommy2d:
I can confirm that the test case works for me.

tvdijen:
I *think* I understand what's going on here.. What happens if you set the `element`-property to `public` instead of the current `protected`?

tommy2d:
It doesn't change a thing I'm afraid. Same error, even after cleaning the sqlite db.

tvdijen:
I don't see what a database has to do with serializing xml, but unfortunately that leaves me with no clues again..

tommy2d:
The serialized attributeValue is stored in sqlite (SimpleSAMLphp_kvstore) right?

none-working (stock) version:

[...]

PR #299, opened 2022-06-26
Fix encrypted id serialization

bartlangelaan:
This fixes the issue from #295, and adds a test to (hopefully) assure that this doesn't happen again.

I'm not 100% sure that this test tests everything, but re-adding the methods makes the test fail so it at least makes sure that it is working better then it currently is.

This _does_ re-introduce some of the issues that were fixed in https://github.com/simplesamlphp/saml2/pull/292, but because it is only a deprecation and not a complete fail, I think it is preferred to release this even if the 'real fix' is not available yet.

### DIFF
--- a/src/SAML2/XML/saml/AttributeValue.php
+++ b/src/SAML2/XML/saml/AttributeValue.php
@@ -156,38 +156,4 @@ class AttributeValue implements Serializable
         $this->element = $doc->documentElement;
     }
 
-
-
-    /**
-     * Serialize this XML chunk.
-     *
-     * This method will be invoked by any calls to serialize().
-     *
-     * @return array The serialized representation of this XML object.
-     */
-    public function __serialize(): array
-    {
-        return [$this->element->ownerDocument->saveXML($this->element)];
-    }
-
-
-    /**
-     * Unserialize an XML object and load it..
-     *
-     * This method will be invoked by any calls to unserialize(), allowing us to restore any data that might not
-     * be serializable in its original form (e.g.: DOM objects).
-     *
-     * @param array $vars The XML object that we want to restore.
-     */
-    public function __unserialize(array $serialized): void
-    {
-        $xml = new self(
-            DOMDocumentFactory::fromString(array_pop($serialized))->documentElement
-        );
-
-        $vars = get_object_vars($xml);
-        foreach ($vars as $k => $v) {
-            $this->$k = $v;
-        }
-    }
 }
--- a/tests/SAML2/XML/saml/AttributeValueTest.php
+++ b/tests/SAML2/XML/saml/AttributeValueTest.php
@@ -116,4 +116,27 @@ ATTRIBUTEVALUE
         $av4->unserialize($ser);
         $this->assertEquals("urn:collab:person:surftest.nl:example", $av4->getString());
     }
+
+
+    /**
+     * Serialize an EncryptedID and unserialize that again.
+     * @return void
+     */
+    public function testSerializeEncryptedID() : void
+    {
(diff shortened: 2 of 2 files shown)


tvdijen:
Hi @bartlangelaan ! Reverting the serialization-patch is not a solution.
Deprecation-warnings are treated as errors in PHP8+, so they will break hard..

bartlangelaan:
Are you sure that this is a change with PHP8? Because that would, like, completely make deprecations unusable. Then every deprecation would be a breaking change.

Deprecations are meant to be warnings until the next major version (PHP9), where all warnings will become errors, right?

tvdijen:
See: https://stitcher.io/blog/new-in-php-8#default-error-reporting-level

bartlangelaan:
Ah, I didn't know that. 

From what I found, it will now report the warnings by default. But that doesn't mean that it is an error that 'will break hard'.
Also, 'display_errors' should only be _on_ during development. So it shouldn't affect any end-user if a deprecation is hit.

So, imho, introducing a deprecation is less worse then breaking all eToegang (eHerkenning / eIDAS) connections and other IDPs that use EncryptedIDs.

tvdijen:
They are treated as errors and will simply break.. Go ahead and try if you don't believe me.
Merging this might fix PHP 8.0 support, but will eventually break in 8.1 because we need the magic methods there.
We just have to _properly_ fix this before we can support PHP 8.x at all..

bartlangelaan:
> They are treated as errors and will simply break.. Go ahead and try if you don't believe me.

https://onlinephp.io/c/c403c

But yes, the magic methods are a requirement for PHP 9.

tommy2d:
Any updates on this issue @tvdijen? Is there anything I can do to assist here?

tvdijen:
No update.. This is nowhere on my radar right now..
I see a lot of issues with supporting PHP 7.4 and 8.1 (8.0 is okayish) simultaneously.. Not just here, unfortunately..

What you could do is dive deeper than I already did and _actually_ find the issue here so we can fix it.
Alternatively you could consider [donating](https://nlnet.nl/donating/) to the project so we can hire a new [maintainer](https://simplesamlphp.org/developers/maintainer).

tvdijen:
@bartlangelaan & @tommy2d could you guys run a test with 582a1fdeca9a676dd58c8ea451cf9235d6be8393 ? I think I fixed it

tommy2d:
When i cherrypick this sole commit, clean all sqlite tables and start a new tvs session in an incognito chrome tab I still get this error:

Uncaught TypeError: SAML2\XML\saml\AttributeValue::getElement(): Return value must be of type DOMElement, null returned in tvs/vendor/simplesamlphp/saml2/src/SAML2/XML/saml/AttributeValue.php:74

Is this to be expected while still being on v4.6.3 of SAML2 and 1.19.5 of simplesaml?

tommy2d:
Same seems to happen when I revert to simplesaml 1.9.7 and the bundled version of saml2, so I'm afraid we're not quite there yet.

tvdijen:
@tommy2d Try this one; 04d363e14140ee5bc51338565f847e639c53e1c1

tommy2d:
Same situation, same error on same line. Just curious: How come the commit date of this commit is in the past?

tvdijen:
Because I squashed the commit with the one from yesterday.
I can't figure out how the element-property can be null :(

tommy2d:
I put a var_export() of $parameters['RawAttributes'] in session_unserialize() and it gives me this:

2022/11/08 10:29:26 [error] 8378#8378: *2707139 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught Exception: array (
  'urn:nl-eid-gdi:1.0:ActingSubjectID' => 
  array (
    0 => 
    SAML2\XML\saml\AttributeValue::__set_state(array(
       'element' => NULL,
    )),
  ),
)

tvdijen:
This is so hard to debug, not being able to reproduce this myself.. I'm starting to think that we have two individual issues here. 1) the serialization of AttributeValue that I supposedly fixed in the referred commit and 2) serialization of a session holding a complex AttributeValue...

Would you know if TVS has some kind of test-environment that is generally available?

tommy2d:
I'm quite certain that there is no general testing environment that you can just hook on to. It's all quite regulated. Even if such a thing were available, we still have a special kind of account ("Meervoudige aansluiting") that is technically different from the regular thing.

I can share my serialized session content with you over the mail. Furthermore, we can also meet here at the office in Hoofddorp/Amsterdam, or at your offices to debug the situation using our test-environment. Please send me a PM / e-mail if you're interested.

tvdijen:
@tommy2d Just to make things 100% sure;  what happens if you update to saml2 v4.6.5 and **clear** all the existing sessions from the sessions table?

tommy2d:
I removed the entire SQLite DB, replaced the composer saml2 folder with the downloaded zip from github releases and I get this error when trying to log on:

PHP message: PHP Fatal error:  Uncaught TypeError: SAML2\XML\saml\AttributeValue::getElement(): Return value must be of type DOMElement, null returned in /tvs/vendor/simplesamlphp/saml2/src/SAML2/XML/saml/AttributeValue.php:73
Stack trace:
#0 /tvs/vendor/simplesamlphp/simplesamlphp/lib/SimpleSAML/Session.php(242): SAML2\XML\saml\AttributeValue->getElement()
#1 [internal function]: SimpleSAML\Session->unserialize()
#2 /tvs/vendor/simplesamlphp/simplesamlphp/lib/SimpleSAML/Store/SQL.php(337): unserialize()
#3 /tvs/vendor/simplesamlphp/simplesamlphp/lib/SimpleSAML/SessionHandlerStore.php(55): SimpleSAML\Store\SQL->get()
#4 /tvs/vendor/simplesamlphp/simplesamlphp/lib/SimpleSAML/Session.php(355): SimpleSAML\SessionHandlerStore->loadSession()
#5 /tvs/vendor/simplesamlphp/simplesamlphp/lib/SimpleSAML/Session.php(270): SimpleSAML\Session::getSession()

tvdijen:
> Is this to be expected while still being on v4.6.3 of SAML2 and 1.19.5 of simplesaml?

Are you still running 1.19.5 ? Because relevant changes have been done in 1.19.6 ...

tvdijen:
I have just released 1.19.7 that has all the fixes in one place. No messing around with vendor-directories and zip-files... It would be great of you could run a test with that one (don't forget to clear the sessions again)

tommy2d:
@tvdijen I can confirm that 1.19.7 fixes the problem! I now however run into a new program with the recently fixed verify_peer for the SAML HTTP Artifact binding. SOAPClient.php seems to assume the peer should be verifiable through the certificate that is used fore signing. But in case of the pre-production tvs, the signing certificate that is included in the IDP metadata (CN: preprod-saml-tvs-rd.dictu.nl) does not equal the certificate offered during the backchannel call (CN: artifact-pp2.toegang.overheid.nl). Both do actually share the same issuer: QuoVadis PKIoverheid Private Services CA - G1.

I worked around this for now by commenting out the verify_peer part, shall I make a separate issue post for the above issue, or do I need to explicitly set a config flag for this to work?

tommy2d:
Hallelujah for you in this case, for fixing the issue ;).

I do understand that this fine grained control is not going to land anymore in 1.19. But how to proceed now? I still need to fork saml2 for it to work with TVS. The previous situation (=no peer verification) worked better than what we have now. Also, I would not mind doing actual peer verification through locally installed CA certificates, even in this case in which a private CA is used for the certificate that is being verified. This however, would also imply changing the current setup in which the signing certificate is assumed to be also used for the https context of the artefact binding.

tvdijen:
I'm going to discuss this internally and will hopefully have a plan for it by the end of the day!

PR #305, opened 2022-08-27
Add interface + trait to be used on several extension points

tvdijen:


tvdijen:
Note should be implemented on md:UnknownRoleDescriptor as well..

src/SAML2/XML/ExtensionPointInterface.php
```diff
@@ -0,0 +1,44 @@
...
+     */
+    public static function getNamespacePrefix(): string;
+
+
+    /**
+     * Return the xsi:type value corresponding this element.
+     *
+     * @return string
+     */
+    public static function getXsiType(): string;
```
jaimeperez, line 43: Shouldn't we get rid of this one? It's kind of redundant now, since we have methods to get both the prefix and the CName separately, and then a bad implementation could lead to inconsistencies.

src/SAML2/XML/ExtensionPointTrait.php
```diff
@@ -0,0 +1,84 @@
...
+{
+    /**
+     * Get the local name for the element's xsi:type.
+     *
+     * @return string
+     */
+    public static function getXsiTypeName(): string
+    {
+        Assert::true(
+            defined('static::NS_XSI_TYPE_NAME'),
```
jaimeperez, line 26: Nitpick: the constants here should probably drop the `NS_` prefix. That made sense for the other set of constants, but in this case they're just used for the XSI type.

src/SAML2/XML/saml/BaseID.php
```diff
@@ -8,91 +8,52 @@
...
         $this->setSPNameQualifier($SPNameQualifier);
     }
 
 
-    /**
-     * Get the type of this BaseID (expressed in the xsi:type attribute).
-     *
-     * @return string
-     */
-    public function getType(): string
```
jaimeperez, line 64: I think we need to keep this one. It's the way we have to retrieve the value we got in the `xsi:type` attribute.

src/SAML2/XML/ExtensionPointInterface.php
tvdijen: Fine with me.. I'll drop it

src/SAML2/XML/ExtensionPointTrait.php
tvdijen: Will perform a search & replace here

src/SAML2/XML/saml/Condition.php
```diff
@@ -8,61 +8,62 @@
...
+
+        $qname = explode(':', $type, 2);
+        if (count($qname) === 2) {
+            list($prefix, $element) = $qname;
+        } else {
+            $prefix = null;
+            list($element) = $qname;
+        }
+        $ns = $xml->lookupNamespaceUri($prefix);
+        $handler = Utils::getContainer()->getElementHandler($ns, $element);
```
jaimeperez, line 61: This one is wrong, we should use `getIdentifierHandler()` instead (we have that method precisely for this use case in particular).

src/SAML2/XML/saml/BaseID.php
tvdijen: It was replaced by getXsiType on the trait, but that one you want removed... So now you got me confused

src/SAML2/XML/saml/Condition.php
tvdijen: The `getIdentifierHandler()` only makes sense on BaseID, not on Condition?

src/SAML2/XML/saml/BaseID.php
```diff
@@ -112,30 +73,37 @@ public function getEncryptionBackend(): ?EncryptionBackend
...
+        $ns = $xml->lookupNamespaceUri($prefix);
+        $handler = Utils::getContainer()->getElementHandler($ns, $element);
 
-        return new self(
-            $type,
-            trim($xml->textContent),
-            self::getAttribute($xml, 'NameQualifier', null),
-            self::getAttribute($xml, 'SPNameQualifier', null)
-        );
+        Assert::notNull($handler, 'Unknown BaseID type `' . $type . '`.');
```
jaimeperez, line 103: We would need to support unknown BaseIDs here.

src/SAML2/XML/saml/BaseID.php
```diff
@@ -8,91 +8,52 @@
...
+use function count;
+use function explode;
 
 /**
  * SAML BaseID data type.
  *
  * @package simplesamlphp/saml2
  */
-class BaseID extends AbstractSamlElement implements BaseIdentifierInterface, EncryptableElementInterface
+class BaseID extends AbstractSamlElement implements
```
jaimeperez, line 29: Shall we make it abstract?

tvdijen: Depends; if we want to instantiate unknown BaseID from the :fromXML(), it cannot be abstract.
If we go with a separate UnknownBaseID class, we can do it..  The constructor is already protected, so it cannot be instantiated immediately

src/SAML2/XML/saml/Attribute.php
```diff
@@ -20,7 +20,7 @@
  *
  * @package simplesamlphp/saml2
  */
-class Attribute extends AbstractSamlElement implements IdentifierInterface, EncryptableElementInterface
+class Attribute extends AbstractSamlElement implements XMLElementInterface, EncryptableElementInterface
```
tvdijen, line 23: This is already implemented on a parent of AbstractSamlElement..

src/SAML2/XML/saml/Condition.php
jaimeperez: You're absolutely right!

tvdijen:
Merging this as-is because it becomes hard to read because it's so big.. Will create a new PR for the md:RoleDescriptor extension point

PR #306, opened 2022-08-29
Refactor Roledescriptor using ExtensionPointInterface

tvdijen:
Continuation of #305


tvdijen:
Strangely this works for other ExtensionPoints like Condition and Statement, but fails on this one. I can move around the namespaces and attributes around as much as I like, but the tests will always have the exact opposite result.

PR #307, opened 2022-08-29
Refactor BaseID/NameIDTypeBaseidtype

tvdijen:
This PR brings back the AbstractBaseIDType that is a mutual ancestor between BaseID and NameIDType.
It also fixes the fact that we could encrypt Issuer, because encryption was implemented on the NameIDType rather than on the NameID.


tvdijen:
Can't seem to figure out why the order of the BaseID-attributes is unstable, causing the last PHPunit error..

PR #314, opened 2022-11-22
Fix SOAP binding Content-Type and SOAPAction processing

tvdijen:
Removed incorrect application/soap+xml;  this is SOAP 1.2 specific and SAML2 doesn't support that


tvdijen:
I take your word for it.. Haven't been able to test it

thijskh:
Verified in a test that it needs to be `HTTP_SOAPACTION`, indeed conforming [with the CGI spec](https://datatracker.ietf.org/doc/html/rfc3875#section-4.1.18)

PR #315, opened 2022-12-15
Remove base64 encoded attributes transformer.

thijskh:
This would decode all received attribute values with base64. As far as I'm aware this is from the early days of federation. In any case SimpleSAMLphp (although it has/had a similar feature) does not use this code. This major release seems like a fine moment to drop this unless someone steps forward with an actual use case.


tvdijen (approved):

tvdijen:
Really confused with the broken tests though... Seems as if Github Actions has changed something

PR #316, opened 2023-01-23
Mutually exclude ACS index / url / ProtocolBinding

tvdijen:
As per spec paragraph 3.4.1;

AssertionConsumerServiceURL [Optional]
[..]
This attribute is mutually exclusive with the AssertionConsumerServiceIndex attribute and is typically accompanied by the ProtocolBinding attribute.

ProtocolBinding [Optional]
[..] This attribute is mutually exclusive with the AssertionConsumerServiceIndex attribute
and is typically accompanied by the AssertionConsumerServiceURL attribute.

### DIFF
--- a/src/SAML2/XML/samlp/AuthnRequest.php
+++ b/src/SAML2/XML/samlp/AuthnRequest.php
@@ -170,10 +170,17 @@ class AuthnRequest extends AbstractRequest
             . ' please specify one or the other.',
             ProtocolViolationException::class,
         );
+        Assert::oneOf(
+            null,
+            [$protocolBinding, $assertionConsumerServiceIndex],
+            'The ProtocolBinding and AssertionConsumerServiceIndex are mutually exclusive;'
+            . ' please specify one or the other.',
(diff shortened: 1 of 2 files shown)


tvdijen:
Tim:

Personally, I think we can be a little bit strict on what we accept here.. I mean, the specs leave no room for debate on this one

Jaime:
I’d like to know first if that could break any major sp software.
It would also be nice to know the reason to make them exclusive
as in… basically confirm that the reason is to avoid ambiguous behavior in the IdP
because if that’s the reason, I think it’s a bit too strict to fail on this… but if there are security implications…

thijskh:
I indeed think it's predictability; otherwise it's undefined which one will be chosen. I would be in favour of making it more predictable if at all possible. But indeed this is useless if it breaks major other implementations. Maybe we can check some logs to see what we can find out about this?

thijskh:
I've checked 15 million logins since 2023-01-01 and 0 of them would violate these assertions.

PR #324, opened 2023-02-09
Registration authority consistency (simplesamlphp#1758)

ghalse:
Change the RegistrationInfo toArray() and fromArray() methods to match SSP as at simplesamlphp#1758

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -16,6 +16,10 @@ the newly added NameIDPolicy-class.
 The API of the StatusResponse-class has been slightly changed; getStatus/setStatus will now handle Status-objects
 instead of the previous array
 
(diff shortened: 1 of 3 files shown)


thijskh (approved):

ghalse:
I was fixing this in SSP rather than here. It was just done here for completeness/consistency with SSP. So far as I can tell, the toArray/fromArray aren't actually (yet) used in SSP. I only proposed this for completeness. I'd be happy to simply close this leaving it as-is, and fix the problem in SSP itself (if by making it more robust).

tvdijen:
That's right, current SSP uses saml2 v4.x and will use that for a while (probably until SSP 3.0).

PR #339, opened 2023-05-31
Include missing import

akinspe:


thijskh (approved):
Thanks, I'll also fix HTTPRedirect.php in the same way.

PR #340, opened 2023-06-02
Fix AttributeValue to deal with complex content

tvdijen:
First of all, this PR fixes the saml:AttributeValue so it will properly deal with complex content.
Second of all a slightly modified version of the EncryptedElementTrait was added to be able to deal with specific requirements for encrypted elements in the SAML spec.


src/SAML2/XML/saml/AttributeValue.php
```diff
@@ -31,25 +37,13 @@ class AttributeValue extends AbstractSamlElement
      *  - string
      *  - int
      *  - null
-     *  - \SimpleSAML\XML\AbstractElement[]
```
tvdijen, line 34: I believe the array was a mistake. It doesn't make sense to add multiple structures in one value if you can add multiple AttributeValues. Also, the default for minOccurs/maxOccurs in XML in `1` and that should imo also count for xs:anyType

src/SAML2/XML/saml/AttributeValue.php
```diff
@@ -106,13 +100,27 @@ public static function fromXML(DOMElement $xml): static
...
         Assert::same($xml->namespaceURI, AttributeValue::NS, InvalidDOMElementException::class);
 
-        $value = $xml->textContent;
-        if (
+        if ($xml->childElementCount > 0) {
+            $node = $xml->firstElementChild;
+
+            if (str_contains($node->tagName, ':')) {
+                list($prefix, $eltName) = explode(':', $node->tagName);
+                $className = sprintf('\SimpleSAML\SAML2\XML\%s\%s', $prefix, $eltName);
```
tvdijen, line 108: I was thinking about creating a utility method to look up known classes, so we can re-use it elsewhere. we shouldn't be using Chunks if we know the element, but we still do that in many places in the lib. The utility method could also query the class-registry to be able to recognize classes outside this lib.

src/SAML2/XML/saml/EncryptedID.php
```diff
@@ -57,8 +42,6 @@ public function decrypt(EncryptionAlgorithmInterface $decryptor): ElementInterfa
         switch ($id) {
             case NameID::NS . ':NameID':
                 return NameID::fromXML($xml);
-            case Issuer::NS . ':Issuer':
```
tvdijen, line 60: Issuer used to extend from NameID in the past, which was a mistake. Issuer is not supposed to be encryptable.

src/SAML2/XML/saml/AttributeValue.php
jaimeperez: It might be. The way I always understood the spec, you could have whatever in there, and that includes XML with multiple elements without a clear root (except for the `AttributeValue` element itself, that is). Not sure if that would make much sense since the idea of putting XML there would be that it works standalone, rather than depending on the `AttributeValue` where it is contained. And I definitely agree that it doesn't make much sense to have multiple elements inside a single `AttributeValue`, when you can just have multiple `AttributeValue` elements, one per each child.

So even if my understanding of the spec is right, I agree that it doesn't make much sense to support this scenario, specially if it makes our life harder.

src/SAML2/XML/saml/AttributeValue.php
jaimeperez: Didn't we create a registry in the container for custom handlers for given XML elements? In any case, I agree we should do that everywhere instead of just using `Chunk`.

src/SAML2/XML/saml/AttributeValue.php
```diff
@@ -106,13 +100,27 @@ public static function fromXML(DOMElement $xml): static
...
+            } else {
+                $value = Chunk::fromXML($node);
+            }
+        } elseif (
             $xml->hasAttributeNS(C::NS_XSI, "type") &&
             $xml->getAttributeNS(C::NS_XSI, "type") === "xs:integer"
         ) {
             // we have an integer as value
-            $value = intval($value);
+            $value = intval($xml->textContent);
```
jaimeperez, line 123: D'oh! Was this a bug?

src/SAML2/XML/saml/AttributeValue.php
```diff
@@ -155,21 +152,28 @@ public function toXML(DOMElement $parent = null): DOMElement
...
                 break;
-            case "array":
-                foreach ($this->getValue() as $object) {
-                    $object->toXML($e);
+            case "object":
+                $value = $this->getValue();
+                if ($value instanceof Chunk) {
+                    // The object is unknown to this library - add an xsi:type for it
+                    $e->setAttributeNS('http://www.w3.org/2000/xmlns/', 'xmlns:xsi', C::NS_XSI);
+                    $e->setAttributeNS(C::NS_XSI, 'xsi:type', $value->getQualifiedName());
```
jaimeperez, line 170: Is the qualified name appropriate as `xsi:type`? I'm basically wondering if we should try to use the `xsi:type` coming from the element itself.

Also, I totally forgot, but does `getQualifiedName()` get you the name with the full namespace, or just the prefix?

PR #343, opened 2023-06-26
Reintroduce container instantiation

tadasrimkus:
It was removed when moving from previous version,
but without it exception is always thrown because `ContainerSingleton` is never instantiated.

I was working on Drupal 10 integration of simplesaml and I had to make a custom patch to resolve saml2 container issue.
Link to `drupal/simplesamlphp_auth` D10 update issue https://www.drupal.org/project/simplesamlphp_auth/issues/3349278

### DIFF
--- a/src/SAML2/Compat/ContainerSingleton.php
+++ b/src/SAML2/Compat/ContainerSingleton.php
@@ -17,6 +17,9 @@ class ContainerSingleton
      */
     public static function getInstance(): AbstractContainer
     {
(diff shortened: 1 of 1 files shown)


tvdijen:
Why are you developing against an alpha-version? It's still a work in progress..
Anyway, the container was [moved](https://github.com/simplesamlphp/simplesamlphp/blob/simplesamlphp-2.0/src/SimpleSAML/Compat/SspContainer.php) to SimpleSAMLphp.

PR #348, opened 2023-10-20
Fix bindings

tvdijen:
- We cannot sign messages without knowledge of the keys. Leave it up to the implementation.
- Move RelayState from messages to bindings


tvdijen:
I no longer support this PR.. Cherry-picked the move for RelayState and closing now

PR #349, opened 2023-11-23
Feature/static calls patch

johanvanmeel:


johanvanmeel:
Remove static calls

johanvanmeel:
Yeah I can understand, this was supposed to be pushed to my own fork to create a quick patch for our environment. The latest version gave us some issues that we need to solve first. I'll close this one, sorry for any inconvenience.

PR #369, opened 2024-12-09
Added explicit nullable types to satisfy PHP 8.4 deprecation

akinspe:
This PR fixes a variety of deprecation warnings found when running in PHP 8.4. 

In particular it fixes "Implicitly nullable parameter types are now deprecated." by explicitly declaring nullable parameters.

### DIFF
--- a/src/SAML2/ArtifactResolve.php
+++ b/src/SAML2/ArtifactResolve.php
@@ -25,7 +25,7 @@ class ArtifactResolve extends Request
      *
      * @param \DOMElement|null $xml The input assertion.
      */
-    public function __construct(DOMElement $xml = null)
+    public function __construct(?DOMElement $xml = null)
     {
         parent::__construct('ArtifactResolve', $xml);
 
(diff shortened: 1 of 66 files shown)


tvdijen:
Thanks!  But I'm not sure we want to merge this. It's a change of signatures and we're already working on a complete rewrite in a new major version..

akinspe:
That's fair. Technically the signatures aren't different as PHP handles the  `type $param= null` explicitly as a nullable type `?type $param=null` . It just means that people will see MANY deprecation warnings in PHP 8.4. Given you had started to support 8.4 in a previous commit I thought you would want to fix this. Here's some more info:

https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated

Thanks for your consideration.

tvdijen:
Oh, since it's completely backwards compatible, I have no problem merging this!

tvdijen:
Actually, the unit tests speak for themselves..  I'll tag a new release

PR #414, opened 2026-07-01
refactor/soap-ssl-validator-openssl

ioigoume:
Remove xmlseclibs coupling by switching SOAP/Artifact validators to OpenSSL-based key extraction and updating tests

### DIFF
--- a/src/Binding/HTTPArtifact.php
+++ b/src/Binding/HTTPArtifact.php
@@ -24,7 +24,6 @@ use SimpleSAML\SAML2\XML\samlp\ArtifactResolve;
 use SimpleSAML\SAML2\XML\samlp\ArtifactResponse;
 use SimpleSAML\Store\StoreFactory;
 use SimpleSAML\Utils\HTTP;
-use SimpleSAML\XMLSchema\Type\AnyURIValue;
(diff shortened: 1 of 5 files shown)


src/Binding/HTTPArtifact.php
```diff
@@ -218,4 +228,70 @@ public function setSPMetadata(Configuration $sp): void
...
+            throw new Exception('ArtifactResponse must be signed.');
+        }
+
+        $keys = $idpMetadata->getPublicKeys('signing', true);
+        if (empty($keys)) {
+            throw new Exception('No signing keys found in IdP metadata.');
+        }
+
+        $alg = $artifactResponse->getSignature()->getSignedInfo()->getSignatureMethod()->getAlgorithm();
+        $signatureMethod = $alg instanceof AnyURIValue ? $alg->getValue() : (string) $alg;
```
tvdijen: What is the idea here? Isn't the algorithm always one of AnyURIValue?

ioigoume: @tvdijen 
It was meant as a defensive fallback in case the xml-security return type changed in the future.
However, since this value is part of signature verification, it’s safer to fail fast on unexpected types rather than silently cast and potentially mask a breaking change

